PR #41, opened 2016-07-27
fix: avoid write after end errors

marionebl:
This fix aims at avoiding the errors related to b75b4fb.
The idea behind this is to ensure `state.cssStream` is always writable,
even when `extract-css` calls `WriteableStream.prototype.end` on it.

---
Disclaimer: I am not that experienced with node streams and thus don't know if this is the best way to do an "infinite" PassThrough; I'd be glad to get some hints / ideas on this.

### DIFF
--- a/handler-js.js
+++ b/handler-js.js
@@ -79,6 +79,7 @@ function wreq (state, bundler, startFn) {
 
   update()
   bundler.on('update', update)
+  state.cssStream.on('finish', onCssStreamFinish)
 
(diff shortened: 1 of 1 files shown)


marionebl:
Any news on this?

yoshuawuyts:
This is great! :grin:

yoshuawuyts:
Released as `v2.0.6` - thank you so much! :sparkles:

PR #47, opened 2016-08-05
Docs: Indicate css method uses sheetify

timwis:
I was debugging for a while trying to figure out why `bankai.css({ use: [ 'sheetify' ] })` wasn't working. On inspecting the code it's clear that `sheetify` is already part of the mix, and `use` is just for transforms on top of sheetify. Hopefully this doc update makes that more clear.


yoshuawuyts:
Nice one, thanks!

Issue #54, opened 2016-08-11
Change --optimize to another word


timwis:
I'd like to suggest changing the `--optimize` flag to another word so that we can use `-o` for the output path in the build tool just like `browserify` offers. Perhaps `--production`?

yoshuawuyts:
haha, but then `--production` would conflict with `--port`. We could totes have `-o` and `-O` flags for optimize and output respectively, you think? - or do you know of a better word?

yoshuawuyts:
there's also the option of adding the `--enhance` flag... https://www.youtube.com/watch?v=Vxq9yj2pVWk

marionebl:
Isn't the `-d/--directory` proposed in #57 the exact same thing as the new meaning of `-o/--out`? `-d` looks just fine to me, let's avoid more option renaming – API stability and documentation is more important than getting the perfect keywords in line imo.

yoshuawuyts:
yeah sounds good

timwis:
sold

PR #57, opened 2016-08-13
Build

timwis:
Ideally `optimize` would be an option, but since it doesn't actually minify/uglify/unassertify the code right now anyway, I figure setting it to `true` in order to disable watchify is good enough for now.

EDIT: Also, does `export` or `write` make more sense than `build`? meh, not a big deal


yoshuawuyts:
LGTM - added you as a collaborator on github and npm, feel free to merge and publish :sparkles:

timwis:
Thanks! This technically breaks a flag introduced by `bankai start`: `--html.entry` is now `--html.script`. Does that warrant a major version bump? Or is it recent enough not to be a big deal?

The only other thing is whether we want to address #54 first. Using `-O` (capital o) for `--optimize` and `-o`/`--output` (like browserify) for `build`.

README.md
```diff
@@ -74,7 +74,7 @@ $ node ./bin/ --help
       -p, --port=<n>         Bind bankai to <n> [default: 1337]
       -o, --optimize         Optimize the page and all assets served by bankai [default: false]
       -b, --browse=<app>     Browse the page served by bankai with <app> [default: false]
-      --html.entry=<uri>     Serve client js at <uri> [default: bundle.js]
+      --html.script=<uri>     Serve client js at <uri> [default: bundle.js]
```
marionebl: Is this really warranted? The breaking change could be avoided easily by transforming `htmlOpts` to fit `create-html` API at [handler-html#L19](https://github.com/yoshuawuyts/bankai/pull/57/commits/3efb2aaac102eed5e21617dd8d9413daa03309fd#diff-e4027ddeefb3dca506057f536fd20752R19)

yoshuawuyts: yup, fair enough

timwis:
@marionebl @yoshuawuyts see revised code to support `--html.entry`. The alternative is to just add `htmlOpts.entry = htmlOpts.script` but then it's in there twice, and I worry about the fact that it's passed off to the global state, so I figured it'd be best to just abstract the back-compat logic immutably. Thoughts? Is there a better way to maintain this compatibility?

bin/build.js
```diff
@@ -0,0 +1,64 @@
...
+      css().pipe(fs.createWriteStream(cssPath))
+    }
+
+    if (html) {
+      const htmlPath = path.join(outputDir, 'index.html')
+      html().pipe(fs.createWriteStream(htmlPath))
+    }
+  })
+
+  callback()
```
yoshuawuyts: hmm, so this callback doesn't really do anything in this case; as it'll most likely be called before the bundles are done resolving. 

Something we could do is wrap the streams in a function, push the functions to an array and resolve them in parallel, calling the callback when done. Something like this:
```js
const parallel = require('run-parallel')
const pump = require('pump')

function build (opts, cb) {

  /* whole
   * lot
   * of
   * stuff
   */

  mkdirp(outputDir, (err) => {
    const fns = []

    if (js) {
      fns.push(function (done) {
        const jsPath = path.join(outputDir, settings.html.js || 'bundle.js')
        pump(js, fs.createWriteStream(jsPath), done)
      })
    }

    if (css) {
      fns.push(function (done) {
        const cssPath = path.join(outputDir, settings.html.css || 'bundle.css')
        pump(css, fs.createWriteStream(cssPath), done)
      })
    }

    if (html) {
      fns.push(function (done) {
        const cssPath = path.join(outputDir, 'index.html')
        pump(html, fs.createWriteStream(htmlPath), done)
      })
    }

    parallel(fns, cb)
  })
}
```
You reckon?

handler-html.js
```diff
@@ -11,28 +10,24 @@ function html (state) {
...
-function createMetaTag () {
-  var metaTag = '<meta name="viewport"'
-  metaTag += 'content="width=device-width, initial-scale=1">'
-
-  return hyperstream({
-    head: { _appendHtml: metaTag }
-  })
+// v3.0 uses --html.entry, but create-html expects `script`
+function createHtmlCompat (opts) {
+  const optsCopy = xtend({}, opts)
```
yoshuawuyts: `xtend(opts)` also works; no need to change but figured I'd point it out C:

bin/build.js
```diff
@@ -0,0 +1,64 @@
...
+// resolve a path according to require.resolve algorithm
+// string -> string
+function resolveEntryFile (relativePath) {
+  const first = relativePath.charAt(0)
+  const entry = ['.', '/'].includes(first) ? relativePath : './' + relativePath
+  return resolve.sync(entry, {basedir: cwd})
+}
+
+function build (options, cb) {
+  const bankai = require('../')({
```
yoshuawuyts: could we move this import statement to the top of the file? - inlining imports like this can be surprising

bin/build.js
```diff
@@ -0,0 +1,64 @@
...
+    if (html) {
+      const htmlPath = path.join(outputDir, 'index.html')
+      html().pipe(fs.createWriteStream(htmlPath))
+    }
+  })
+
+  callback()
+}
+
+module.exports = build
```
yoshuawuyts: Could we move the exports to the top of the file? Generally I find `module.exports` is the first thing I'm looking for when looking at a new file, as it tells me what the entry point is. Having it live at the bottom of the page makes scanning over a file harder :cry:

README.md
```diff
@@ -42,7 +42,7 @@ http.createServer((req, res) => {
 
 ### assets.html(opts?)
 Return an `html` stream. Takes the following options:
-- __opts.entry:__ `js` entry point. Defaults to `/bundle.js`
+- __opts.script:__ `js` entry point. Defaults to `/bundle.js`
```
yoshuawuyts: or umm, would `html.js` make more sense?

yoshuawuyts:
> EDIT: Also, does export or write make more sense than build? meh, not a big deal

Hmh, yeah they all feel quite close to each other. I've only got a feeling to go on but I think that `build` is like a common enough command that it'll be memorable, whereas the others maybe a bit less? Haha yeah tho I'm not too fussed, figure we should just pick one and go with it

---
> so I figured it'd be best to just abstract the back-compat logic immutably

Yeah you're right, we should go for the best UX - the fact that JS and HTML can be configured separately is an implementation detail we should abstract away. What a user wants is to be able to specify the JS entry point, and the CSS entry point and then get all the stuff on their screen / bundled to a dir. Whatever we can do to make this as painless as possible would be best :sparkles:

bin/build.js
timwis: love it!

bin/build.js
timwis: sure thing - got that from `start.js`

README.md
timwis: Shoot, I forgot to remove this documentation change -- I changed it back to `entry` so as not to pose a breaking change. Anyway, not sure what you mean with your question.

timwis:
Thanks for the feedback! Should be all incorporated.

README.md
yoshuawuyts: haha I don't even remember anymore - ah well, glad docs are consistent now! :D

yoshuawuyts:
Cool, LGTM!

PR #59, opened 2016-08-14
fix: resolve js entry from cwd

### LINKED ISSUE #46: JS module requires an absolute path to serve
It appears `bankai.js(browserify, './src/index.js')` throws an error `Cannot find module './src/index.js'`. It only works when I give it the full path, `/Users/tim/...etc`. I think this is because of `require.resolve` on [this line](https://github.com/yoshuawuyts/bankai/blob/master/handler-js.js#L29). Not very familiar with `resolve` but would `path.resolve` be a better fit perhaps? Or am I just using it incorrectly?

marionebl:
*  default `opts.js.basedir` => `browserify.basedir` to `process.cwd()`
*  extend basic example with `npm run build` => `bankai build`
*  add API usage example at basic example
*  showcase `opt.js.basedir = __dirname` usage at basic example
*  document `bankai.js` defaults
*  improve basic example documentation
*  remove mention of HMR (for the time being)

fix #46


yoshuawuyts:
Heya, thanks again for the patch! Hmh, for this to work well we should be able to override the basedir with an option like how `browserify` does it ([link](https://github.com/substack/node-browserify/blob/8f7adc96e3ed59c299d078405a463860f5409ba6/test/entry_relative.js#L29)). If we don't do it that way running `bankai` from within a service (e.g. not started from the same dir as where it executes) it might hit snags. Does that make sense? Thanks!

marionebl:
Makes sense, will adapt the patch accordingly this evening

marionebl:
Updated!

yoshuawuyts:
Sweet, looking good. Added you as a contrib on npm and github. Feel free to release :rocket:

PR #56, opened 2016-08-13
Register js after css

timwis:
Currently, the example throws the error `css must be registered before js to work`

PR #61, opened 2016-08-15
docs: fix "css must be registered before js to work"

juliangruber:


yoshuawuyts:
Yeah you're totes right; this should change soon but better fix examples before that. Thanks!

timwis:
😿 #56  :P

PR #65, opened 2016-08-19
Move some devDeps to deps to support global bin usage

### LINKED ISSUE #64: Can't run from global install
Deps such as `get-server-port` are in `devDependencies` and aren't installed via `npm install -g bankai`. I think browserify too, but gotta run!

timwis:
Fixes #64


timwis:
Verified it works on c9 environment using `npm i -g yoshuawuyts/bankai#64-devdeps-global`

PR #69, opened 2016-08-24
Clean up package.json and add transform example

bcomnes:
This removes unneeded packages from the package.json, and adds a browserify field and es2020 as an example browserify transform to show that bankai's instance of browserify will indeed read from this configuration source.

It also provides an example of using a sheetify transform.


yoshuawuyts:
yeah I like this. Thanks!

PR #71, opened 2016-08-25
start: fix flags

yoshuawuyts:
The `--open` flag was doing weird things; wasn't defined. Moved some exports to the top and helper functions to the bottom. Also pyramids. Sorry not sorry for the bike shed; I'd have a linter for this stuff but writing linters is hard so hah


yoshuawuyts:
`v3.2.0`

PR #73, opened 2016-08-30
correct default opts

f-mer:
[create-html](https://github.com/sethvincent/create-html)  takes _head_ as an option for meta-tags.


yoshuawuyts:
Ah cool, thanks!

PR #74, opened 2016-08-31
fix: ensure node 4 compat

### LINKED ISSUE #66: start doesn't work on node v4.4.5
Just FYI, `const entry = ['.', '/'].includes(first)` throws an error on `includes` as it's an ES6 feature

marionebl:
*  add 'use strict' to all files for node 4 block-scoping
*  remove usage of `Array.prototype.includes` in resolveEntryFile
*  rename `resolveEntryFile` to `resolveEntry`
*  move multiple `resolveEntry` implementations to `lib/resolve-entry`
*  order imports in `bin/build` according to length
*  add lib to allowed files in manifest

fixes #66

### DIFF
--- a/bin/start.js
+++ b/bin/start.js
@@ -1,8 +1,10 @@
+'use strict'
+
+const resolveEntry = require('../lib/resolve-entry')
 const stringToStream = require('string-to-stream')
 const getServerPort = require('get-server-port')
 const serverRouter = require('server-router')
 const browserify = require('browserify')
-const resolve = require('resolve')
 const xtend = require('xtend')
 const http = require('http')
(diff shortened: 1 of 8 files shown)


bin/start.js
```diff
@@ -1,8 +1,10 @@
+'use strict'
+
+const resolveEntry = require('../lib/resolve-entry')
```
marionebl: @yoshuawuyts Are you okay with this? The current project setup seems to suggest you'd like to keep together in very flat organisation.

marionebl:
Going forward with this to fix the issues with node version < 6. We can revisit any stylistic issues later.

PR #78, opened 2016-09-17
Serve bundle.css and bundle.js from absolute path

### LINKED ISSUE #32: figure out how to serve bundles from the right locations
https://github.com/yoshuawuyts/choo/pull/73 was tripping up; look into _why_ this happens and at the very least document this behavior

timwis:
fixes #32, though as I mentioned in that issue I'm not sure whether this would break other use cases (ie. a site served at a subdirectory, `localhost/site-root/`). Just wanted to get conversation going because at the moment you can't serve an app with subroutes with bankai. Thoughts?

### DIFF
--- a/handler-html.js
+++ b/handler-html.js
@@ -12,8 +12,8 @@ function html (state) {
   return function initHtml (opts) {
     opts = opts || {}
     const defaultOpts = {
-      entry: 'bundle.js',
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
I think this looks good - if it solves the issues then I'm cool with it C:

PR #81, opened 2016-09-29
Implement live reload

aknuds1:
Implement live reload through tiny-lr. Also add browser test suite, using Phantom.js.


package.json
```diff
@@ -49,8 +49,10 @@
     "choo": "^3.2.0",
     "concat-stream": "^1.5.1",
     "dependency-check": "^2.6.0",
+    "es2020": "^1.1.9",
```
aknuds1: Add required dependency for running examples.

package.json
```diff
@@ -49,8 +49,10 @@
...
     "is-html": "^1.0.0",
     "istanbul": "^0.4.4",
+    "normalize.css": "^4.2.0",
```
aknuds1: Add required dependency for running examples.

test/index.js
```diff
@@ -87,6 +87,9 @@ test('js', function (t) {
       }))
     })
   })
+
+  t.test('js reloads upon source file modification', function (t) {
```
aknuds1: Add pending test for new functionality.

aknuds1:
How can I test reloading of JS due to source file modification?

timwis:
@aknuds1 perhaps [budo's tests](https://github.com/mattdesl/budo/blob/master/test/test-live.js) would be useful?

aknuds1:
Thanks @timwis, will look at them.

aknuds1:
@timwis The budo test for reload upon source file modification seems to rely on budo sending 'update' and 'reload' events. This seems highly budo specific and I'm not sure how this verifies that the development server actually reloads the code, but anyhow what would be the bankai equivalent??

aknuds1:
From the looks of it, budo uses [tiny-lr](https://www.npmjs.com/package/tiny-lr) to implement live reload. Should I try to implement live reload in bankai through tiny-lr as well?

timwis:
> Basically, I think what I would really want for this test is to verify as closely as possible that bankai ensures that the code is reloaded after source code changes.

For what it's worth, this describes the "watchify" functionality that bankai provides, which would technically be a separate feature from live reload. The watch/rebuild functionality is implemented already, but there don't appear to be any tests for it. 

I'd expect live reloading would be something that's triggered when the watchify functionality finishes (or the css equivalent), and I can imagine that writing a test for that would be slightly different.

I haven't used tiny-lr personally but budo's seems to work well so I don't see why not, unless @yoshuawuyts knows a better alternative. Either way I'd suggest pulling it in via `npm install --save <package-name>`.

Thanks for working on this! I'll definitely enjoy this feature :)

aknuds1:
@timwis Hm, not sure if I understand exactly what you mean, but I'm new to watchify and live reload and all this stuff. Basically, I want to verify live reload functionality as close to the metal as feasible if you see what I mean, in order to have the highest possible assurance that the feature actually works. I'm still not sure how to write this sort of test in the bankai context.

Anyhow, I'll start implementing tiny-lr in bankai, looks as if it requires an HTTP middleware layer judging from budo's code. I'm new to this as well, so let's experiment ;)

yoshuawuyts:
`tiny-lr` is one path that we could take; it can't handle HMR stuff tho - I think the more generic solution would be to send new bundles over a websocket / SSR and then update the appropriate HTML tag (e.g. `<script>` or `<style>`) - believe that's in essence what `tiny-lr` does anyway, but this way we could keep the deps a bit smaller. Ghmmmm

aknuds1:
@yoshuawuyts What do you think about copying budo's approach (i.e. tiny-lr) for now, so we can make it work relatively quickly?

yoshuawuyts:
@aknuds1 sure, go for it! - the way you're adding it with splitting most of it into a separate file means it can later be replaced without too much effort (:

aknuds1:
Hooray! ;)

yoshuawuyts:
@aknuds1 ps just saw you're in berlin and working on this rn - I've got a bike, if you've got time I would be happy to pair and get this out asap haha :tada:

aknuds1:
@yoshuawuyts Yes, we met at Node.js meetup and then Betahaus ;) I'm just very busy the next two days and then I'm going to London for the weekend :( Otherwise I would love to!!

aknuds1:
Berlin housing market is basically killing my ability to get shit done :(

yoshuawuyts:
Ohhh, that's you! Haha, didn't recognize you because of the avatar haha :sparkles: - haha good luck :D :D

aknuds1:
Thanks! I might have found a sublet today! Fingers crossed!!!

aknuds1:
The live reload implementation is looking pretty good now, I even have a test for it! Granted, I'm using the [Nightmare](http://www.nightmarejs.org/) headless browser and I'm not sure how the rest of you feel about this type of testing strategy. It's the best I could come up with however. Nightmare is also proving to be a bit flaky with random test failures unfortunately, so I guess it'll have to be replaced with something more stable in any case.

Also, it should be mentioned that the live reload implementation is less complete than budo's in that css/html files are not generally monitored, only the css files that are bundled by browserify.

aknuds1:
I rewrote the live reload test using [Phantom.js](https://github.com/amir20/phantomjs-node) instead of Nightmare. It works well!

aknuds1:
I would say the branch is ready except for the reduced code coverage, will have to get around to that.

aknuds1:
@yoshuawuyts Regarding code coverage, do we ever get errors [from the browserify stream](https://github.com/yoshuawuyts/bankai/pull/81/files#diff-bba08292d6149727ddfdfaa846d91751R41)? AFAICT, [errorify](https://github.com/zertosh/errorify) prevents us from receiving errors from browserify. I even wrote a test inducing bundling failure through syntax error in the source code, and it does seem that errorify is trapping it.

As such, I wonder if we should drop the above error handling.

yoshuawuyts:
yeah this patch is not gonna merge cleanly - think I'm gonna close this in favor of the approach in https://github.com/yoshuawuyts/bankai/issues/31

PR #83, opened 2016-10-22
Use eslint

aknuds1:
Start using eslint, in order to enforce code standards and help in eliminate errors.

**NOTE**: We should agree on these rules before merging.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,0 +1,25 @@
+parserOptions:
+  ecmaVersion: 7
+env:
+  node: true
+rules:
+  strict:
+    - 2
+    - "global"
+  quotes:
+    - 2
+    - "single"
+    -
+      allowTemplateLiterals: true
+  new-cap: 0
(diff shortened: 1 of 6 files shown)


.eslintrc.yaml
```diff
@@ -0,0 +1,25 @@
+parserOptions:
+  ecmaVersion: 7
```
aknuds1: Which ECMA version do we standardize on?

.eslintrc.yaml
```diff
@@ -0,0 +1,25 @@
...
+  keyword-spacing:
+    - error
+    -
+      after: true
+  space-before-function-paren:
+    - error
+    -
+      anonymous: always
+      # TODO: Decide on rule
+      named: ignore
```
aknuds1: What is the rule for bankai?

aknuds1:
Nevermind, I just realized the project uses the JavaScript Standard Style linter.

PR #84, opened 2016-10-22
Add "lint" npm script

aknuds1:
Add the ability to invoke `npm run lint` for quick linting.

### DIFF
--- a/example/basic/server.js
+++ b/example/basic/server.js
@@ -13,7 +13,7 @@ const routes = [
   ['/', html],
   ['/404', html],
   ['/bundle.css', css],
-  ['bundle.js', js]
+  ['/bundle.js', js]
 ]
 
 const router = serverRouter('/404', routes)
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "deps": "dependency-check . && dependency-check . --extra --no-dev",
     "test": "npm run lint && npm run deps && NODE_ENV=test node test/index.js",
-    "lint": "./node_modules/.bin/standard",
+    "lint": "standard",
     "test:cov": "npm run lint && npm run deps && NODE_ENV=test istanbul cover test/index.js"
   },
   "repository": "yoshuawuyts/bankai",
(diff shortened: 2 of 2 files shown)


example/basic/index.js
```diff
@@ -1,3 +1,4 @@
+'use strict'
```
yoshuawuyts: I'm personally not a fan of `'use strict'` - could we leave this out? Thanks!

package.json
```diff
@@ -8,8 +8,9 @@
   },
   "scripts": {
     "deps": "dependency-check . && dependency-check . --extra --no-dev",
-    "test": "standard && npm run deps && NODE_ENV=test node test/index.js",
-    "test:cov": "standard && npm run deps && NODE_ENV=test istanbul cover test/index.js"
+    "test": "npm run lint && npm run deps && NODE_ENV=test node test/index.js",
+    "lint": "./node_modules/.bin/standard",
```
yoshuawuyts: `npm` automatically injects the local `node_modules/.bin` dir into `$PATH` when calling `npm run <script>` - `"lint": "standard"` should suffice here

example/basic/index.js
aknuds1: Cool, we should just remove it consistently from all source files then? I was actually wondering what we should do on a project wide basis, since the linter doesn't care.

package.json
aknuds1: @yoshuawuyts Thanks for catching this.

example/basic/server.js
```diff
@@ -13,7 +13,7 @@ const routes = [
   ['/', html],
   ['/404', html],
   ['/bundle.css', css],
-  ['bundle.js', js]
+  ['/bundle.js', js]
```
aknuds1: I just noticed that this path ought to start with a slash, at least for consistency.

example/basic/index.js
yoshuawuyts: yeah, think that's best - afaik I haven't added it anywhere, not sure if
someone
else snuck it in tho

yoshuawuyts:
this is also not merging cleanly anymore; closing for now

PR #88, opened 2016-11-14
Readme spelling fix, move a from stream to ssets

joehand:
🅰️ ▶️


marionebl:
Thanks for lending a hand!

PR #96, opened 2016-11-16
feat(deps): bump sheetify

yoshuawuyts:
bump to the latest sheetify. Removes the need for the `insert-css` peer dep. Also makes it so styles are global by default and only `:host` creates prefixes. Kinda feel it's a semver major.

Thoughts?


yoshuawuyts:
yaaaaaaa, merging :sparkles:

yoshuawuyts:
`v5.0.0`

PR #103, opened 2016-11-20
Fix options

aknuds1:
After merging your changes, I see from my tests that ignoring CSS is still not working. This patch should rectify that.


yoshuawuyts:
Merged in b53f6c2 - thanks!

PR #106, opened 2016-11-28
feat(index): ignore electron

yoshuawuyts:
Ignore electron by default so we can bundle full apps for electron and like not have to worry electron's built-ins will be picked up in the bundle


yoshuawuyts:
nah let's not

PR #120, opened 2017-01-22
[WIP] Add support for serving static files

marlun:
This works right now if you have your files under assets or specify another folder with --assets=<foldername> however I'm not sure about the best way to handle if the user adds slashes to the dirname and how to check if we're trying to access a static asset or not.

What would you suggest?


bin.js
```diff
@@ -136,14 +136,13 @@ function start (entry, argv, done) {
   })
 
   function static (req, res) {
-    if (req.url.indexOf('/' + argv.assets + '/') !== 0) {
+    if (req.url.indexOf('/' + argv.assets) !== 0) {
```
yoshuawuyts: Using a regex might be more effective:
```js
var regex = new RegExp('\/' + argv.assets)
if (regex.test(req.url)) {
```

marlun: Changed.

marlun:
Trying to get the 404 to work now. Not sure why it's not working. static currently looks like:

    function static (req, res) {
      var regex = new RegExp('\/' + argv.assets)
      if (regex.test(req.url)) {
        var file = path.join(path.dirname(entry), req.url.substr(1))
        fs.exists(file, function (exists) {
          return exists
            ? assets.static(req, res).pipe(res)
            : (res.statusCode = 404 && res.end('404 not found'))
        })
      } else {
        return (res.statusCode = 404 && res.end('404 not found'))
      }
    }

But when I add a src="..." which doesn't exist I still get 200 OK but it fails :)

yoshuawuyts:
@marlun hmmm, wonder what's up with that - think if it's only the wrong statuscode that's holding this up we could perhaps ignore it and move forward anyway? :p

index.js
```diff
@@ -122,3 +123,8 @@ Bankai.prototype.css = function (req, res) {
     return from([this._css])
   }
 }
+
+// (obj, obj) -> readStream
+Bankai.prototype.static = function (req, res) {
+  return send(req, req.url.substr(1));
```
yoshuawuyts: extra semi

marlun: Fixed. Was fixed locally but I forgot to push the change because I have the following change in index.js too which seems necessary as soon as the url path is not `/` but which I don't want to add in this pull-request 😄 

```diff
diff --git a/index.js b/index.js
index 051e555..9168e87 100644
--- a/index.js
+++ b/index.js
@@ -42,8 +42,8 @@ function Bankai (entry, opts) {
 
   this._html = (function () {
     var base = {
-      script: 'bundle.js',
-      css: (self.cssDisabled) ? null : 'bundle.css',
+      script: '/bundle.js',
+      css: (self.cssDisabled) ? null : '/bundle.css',
       head: '<meta name="viewport" content="width=device-width, initial-scale=1">'
     }
     var html = createHtml(xtend(base, opts.html))
```

yoshuawuyts:
oh actually we should also include a build step for this. e.g. do a [fs.readdirsync](https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options) and append the contents to the array of files to copy. Also make the right output dir and handle the case where there's no input / output assets dir. uehhhh

yoshuawuyts:
`v5.4.0` :tada:

PR #122, opened 2017-01-26
Add option for returning index on unknown URLs

toddself:


toddself:
From #choo:

```
12:46 <marlun> this might be what several people has talked about here before but when using bankai and I go to localhost:8080/site/id I just get 404 Not Found, I need to go to / and then when I programatically change the location to /site/1 the router works. If I reload now though I get 404 again.
13:11 <toddself> marlun: you need to tell bankai to serve the index for all URLs
13:11 <toddself> i believe, by default, it only serves the data from /
13:12 <marlun> yeah I saw that too and the way it works now I believe you can't tell bankai to do that. (?) ...will look into it :)
13:12 <toddself> hmm
13:12 <toddself> i know budo can do it
13:12 <toddself> (that's the tool i use)
13:13 <toddself> but i don't see anything in bankai's doc...
13:14 <rahat> probably needs a PR for pushstate
13:14 <toddself> looks like bin.js
13:14 <toddself> needs a change to line 115
13:15 <toddself> one sec i'll make a PR for it and see if yoshua likes it :-D
13:19 <toddself> https://github.com/yoshuawuyts/bankai/pull/122
13:19 <toddself> marlun: you can try that out by cloning that repo from github, checking out the branch and doing an `npm i -g .` from there
13:20 <toddself> passing in `-a` will return the index file for all unknown URLs
```

rahatarmanahmed:
imo i'd expect this option to be named `--pushstate`

toddself:
I'm not so sure.  Push state is just one way to change the URL the browser is displaying (but doesn't even have to require re-loading something from the server).

The server has no clue you've used push-state. If you wanted to just start bankai and then go to `localhost:8000/foobar` instead of `/` off the bat, this handles that, but calling it push state might not communicate that effectively?

toddself:
But it doesn't really matter what color the bikeshed is :).  The feature is important not the words :)

rahatarmanahmed:
I don't think there is any other reason to return index.html on all paths
other than an app using pushstate. But yeah, not really a big deal.

amsross:
As a new entrant to choo, bankai, browserify, and really this whole toolchain, the fact that the ability to do this wasn't in here tripped me up for about an hour. I ended up looking at budo (which I couldn't get working, but that's a whole different thing) and seeing its use of the `--pushstate, -P` option, so I came back here and looked for similar functionality. Thankfully there aren't many issues or PRs on this repo, because the fact that these two concepts are related isn't readily apparent to the uninitiated without a more-than-cursory read.

TL;DR: Calling it the same thing could be helpful to people like myself who are unfamiliar in this world. But yeah, obvious bikeshedding.

yoshuawuyts:
I reckon it might make more sense to catch `Content-Type="html"` and then always serve up our `index.html`

toddself:
👍

toddself:
Good news everybody!  There is no longer an option for `allhtml` or `pushstate`.  It just does it by default, so there's nothing to name :)

bin.js
```diff
@@ -34,7 +34,8 @@ var argv = subarg(process.argv.slice(2), {
     optimize: 'O',
     port: 'p',
     verbose: 'V',
-    version: 'v'
+    version: 'v',
+    allURLS: 'a'
```
amsross: This isn't used anymore

toddself:
@amsross -- fixed

bin.js
```diff
@@ -34,7 +34,7 @@ var argv = subarg(process.argv.slice(2), {
     optimize: 'O',
     port: 'p',
     verbose: 'V',
-    version: 'v'
+    version: 'v',
```
yoshuawuyts: dangling comma, think this trips up standard

yoshuawuyts:
Just the one syntax thing, good to merge otherwise - you should be a collab so feel free to merge && publish :sparkles:

Issue #136, opened 2017-03-20
bankai@5.5.0: requesting an asset causes crash


noffle:
In this case I was trying to serve `bunny.png` from the current directory. The process crashes with the following after the request for `/bunny.png` is received. This is a regression from `5.4.0` -- I downgraded to that for now, which seems to work fine.

```
[sww@figure8 BoardingAction]$ bankai -a . game.js
17:48:25 ✨  bankai Started for game.js on http://localhost:8080
17:48:25 ✨  http req /
17:48:25 ✨  http res / GET:200 46ms 242B
17:48:25 ✨  http req /bundle.css
17:48:26 ✨  http req /bundle.js
17:48:36 ✨  http res /bundle.css GET:200 10954ms
17:48:36 ✨  http res /bundle.js GET:200 10965ms 1.3 MB
17:48:38 ✨  http req /bunny.png
/home/sww/.npm-global/lib/node_modules/bankai/node_modules/send/index.js:873
  if (this._acceptRanges && !res.getHeader('Accept-Ranges')) {
                                 ^

TypeError: res.getHeader is not a function
    at SendStream.setHeader (/home/sww/.npm-global/lib/node_modules/bankai/node_modules/send/index.js:873:34)
    at SendStream.send (/home/sww/.npm-global/lib/node_modules/bankai/node_modules/send/index.js:632:8)
    at onstat (/home/sww/.npm-global/lib/node_modules/bankai/node_modules/send/index.js:737:10)
    at FSReqWrap.oncomplete (fs.js:112:15)
```

lrlna:
having the same issue actually; gonna dig in

PR #140, opened 2017-03-21
fix assets not serving (#136)

noffle:
2c13e255ce7ffa3fff8e3a8b424d398ffceb472d added response size logging via
`server-sink`, which acts as a passthrough stream between `req` and `res`.

However, `send`, which streams the static files from disk, has a stateful
`pipe()` method and assumes that the destination stream is an
HttpServerResponse. Since the sink through stream is not, things break.

This PR just foregoes response size logging on assets. I'd be happy with just
changing the static stream code to simply

```
fs.createReadStream(req.url.substring(1)).pipe(sink)
```

but we'd lose out on the fancy features of `send` (if they're desired by
others). Modifying `send` itself to be tolerant of intermediary passthrough
streams looks difficult -- there are some pretty deep assumptions in place
about it knowing where the `res` is and being able to mutate it.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -134,7 +134,7 @@ function start (entry, argv, done) {
     } else if (req.headers['accept'].indexOf('html') > 0) {
       assets.html(req, res).pipe(sink)
     } else if (staticAsset.test(req.url)) {
-      assets.static(req, res).pipe(sink)
+      assets.static(req).pipe(res)
     } else {
       res.writeHead(404, 'Not Found')
       sink.end()


yoshuawuyts:
Oh dang - I think this might leak mem though as sink isn't closed. We should probably have a patch in server sink to support send() somehow - e.g. forward all the events. Does that sound reasonable?

yoshuawuyts:
Good for now; figured out to get the res size even if streaming https://github.com/yoshuawuyts/benchmark-anything-server/blob/master/lib/index.js#L24 so we can migrate to this approach later on. Woot!

yoshuawuyts:
`v6.1.1` - thanks heaps! (and sorry for not merging earlier; been sick for the past few days :( )

noffle:
Thanks yosh! I was meaning to get around to making the changes you mentioned, but glad this was Good Enough(tm) for now.

PR #145, opened 2017-03-23
support address option. 

### LINKED ISSUE #144: listen on host different to localhost
This is a litle feature request, but maybe is an edge case. The thing is that listening on localhost works fine for most cases, but in my case, I'm developing on a cloud ide (c9.io) and I can run almost anything with the litle condition that it must be run on `0.0.0.0:8080` and not `localhost:8080`. So, running `bankai start -p 8080` is crashing.
So, would it be posible to pass an option to bankai cli to configure the host? what would be a good option, since `-h` and `-H` are already taken? Maybe `-a` for `--address`?

YerkoPalma:
closes #144

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -28,7 +28,8 @@ var argv = subarg(process.argv.slice(2), {
     debug: false,
     open: '',
     optimize: false,
-    port: 8080
+    port: 8080,
+    address: 'localhost'
   },
   alias: {
     assets: 'a',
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
`v6.1.0` - thanks heaps! - sorry for not merging earlier; I had a stomach flu since Thursday and wasn't online much

YerkoPalma:
It's ok :smile: Hope you are better now.

PR #150, opened 2017-03-29
added envify

### LINKED ISSUE #149: include envify
As per https://github.com/yoshuawuyts/choo/issues/444 (woot, triples!) we should probably be running `envify` to do the `process.env.NODE_ENV` stuff for `--optimize`. I think this should be fairly similar to other transforms. Happy to help if someone wants to pick this up!

YerkoPalma:
fix #149 

Also added an entry on package.json scripts to test it, but I'm not sure if you are ok with a _build_ script for a tests, it sounds to me like it builds the whole module. Tell me if you wanna change it, otherwise is ready to merge.


package.json
```diff
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "start": "./bin.js example.js --open -p 8080 -A 0.0.0.0",
+    "build": "NODE_ENV=production ./bin.js build example.js --optimize",
```
zigomir: Not sure, but wouldn't it be even more handy that one could expect that `--optimize` does that without a need to specify `NODE_ENV=production`?

I prefer explicit over implicit though, but in this use case I'd expect from `--optimize` to do the right thing without a need to specify (and for that matter we'd also need to document it) NODE_ENV. Or is this too magical? I'm not really sure 😊

index.js
```diff
@@ -82,6 +83,7 @@ function Bankai (entry, opts) {
     if (self.optimize) {
       b.transform(unassertify, { global: true })
       b.transform(yoyoify, { global: true })
+      b.transform(envify, { global: true })
```
yoshuawuyts: I don't think we'd need to apply this as a global transform; running `process.env.NODE_ENV` checks in libraries seems like a practice that shouldn't be encouraged / might cause issues

package.json
yoshuawuyts: Hmmm, yeah actually I agree a little bit here; I think it might make sense to always run `envify` so that `process.env.NODE_ENV` is always respected. That way people can always run:
```js
if (process.env.NODE_ENV === 'production') // do foo
else // do bar
```

index.js
yoshuawuyts: As per the comment below it might perhaps even make sense to always run this transform. Hmmmm

YerkoPalma: So, the changes would be

- no global
- set NODE_ENV to production so users don't have to do it themselves

something like

```js
b.transform(envify({
  NODE_ENV: 'production'
}))
```

is that correct? @yoshuawuyts @zigomir

yoshuawuyts: Yes to the first, no to the second:
```js
b.transform(envify({
  NODE_ENV: process.env.NODE_ENV
}))
```
Although I thought `envify` would take care of that out of the box, if you're not using `envify/custom`

YerkoPalma: Oh yeah to do that we would have to use `envify/custom`. Anyway, I'm a bit lost right now haha. `envify` replace env vars but we have to set them in some where, so if we are not setting `NODE_ENV` then there is no way to avoid `NODE_ENV=production` or I'm missing something?

Other thing we could do is maybe define a special bankai env var, something like `BANKAI_ENV` and document it properly

YerkoPalma: @yoshuawuyts any updates on this? should I only remove the global option?

yoshuawuyts:
`v7.2.0` :tada:

PR #151, opened 2017-04-02
add --watch flag

yoshuawuyts:
Semver major. Useful to have `--optimize` available with `bankai start`
so you don't need to keep reloading the servereven if you want optimized
output.


yoshuawuyts:
Closing in favor of #153 which does like wayyy more and kinda depends on this patch. The implementation of `--watch` is buggy here anyway, fixed it in #153

PR #154, opened 2017-04-05
add an ability to watch files on start and build

lrlna:


bin.js
```diff
@@ -89,6 +86,7 @@ function main (argv) {
   if ((argv._[0] !== 'build') && (argv._[0] !== 'start')) {
     argv._.unshift('start')
   }
+  console.log(argv)
```
yoshuawuyts: `console.lol`

lrlna:
As if

PR #156, opened 2017-04-07
Updated js and css default paths

delvedor:
The compiled html files has the js and css path written in this way: `/bundle.*`.
I don't know if others have encountered my same problem, but both chrome and electron throw the following exception: `Failed to load resource: net::ERR_FILE_NOT_FOUND`.

With this fix it will not happen anymore.

Cheers!

ps if I'm doing something wrong, please let me know, maybe is just me :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,8 +49,8 @@ function Bankai (entry, opts) {
 
   function html () {
     var base = {
-      script: '/bundle.js',
-      css: self.cssDisabled ? null : '/bundle.css',
+      script: './bundle.js',
+      css: self.cssDisabled ? null : './bundle.css',
       head: '<meta name="viewport" content="width=device-width, initial-scale=1">'
     }
     var html = createHtml(xtend(base, opts.html))


yoshuawuyts:
yeah this seems like an error on your part; if you server the resulting http from a remote http server I'm pretty sure it will do funky resolving things

delvedor:
I'm using the generated code with electron, and serving it with 
```js
mainWindow.loadURL(url.format({
  pathname: path.join(__dirname, 'build-browser', 'index.html'),
  protocol: 'file',
  slashes: true
}))
```
Probably the cause is that I'm using the `file` protocol, in fact I get the same error if I open the file directly into the browser.

yoshuawuyts:
Oh yeah, don't serve the HTML included in bankai - it breaks horribly once you compile. Just include a static file and leave the `assets.html()` call for what it is haha

delvedor:
Perfect, thanks!

PR #158, opened 2017-04-10
Flag off uglifyify

toddself:
Add a flag to prevent uglifyify from running on the assets to let people run
es6 code more easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@
       -p, --port=<n>          Bind bankai to a port [default: 8080]
       -V, --verbose           Include debug messages
       -w, --watch=<bool>      Toggle watch mode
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
Oh, the way I was thinking to do this would be to set `uglify` as a boolean, and default to `true` - never been a fan of the negate-flags haha. E.g. the API would become:
```sh
$ bankai start --uglify=false index.js
```

yoshuawuyts:
Oh, also this would need to be documented in the `bin.js` help text and `readme.md`

toddself:
Oh good point.  The flag has been changed to be like `watch`, and the README is updated now with a note about ES5 syntax.

yoshuawuyts:
`v7.1.0` :tada:

PR #167, opened 2017-04-22
fix broken uglify switch

lordvlad:
Currently the uglify switch is parsed as a string, and evals true when checked with `if  (opts.uglify)`. This commit fixes that.


yoshuawuyts:
Heya, thanks for the patch! `yo-yoify` should always be run; by not passing it on the code the runtime execution becomes 2-3x as slow on average. I'm :+1: on fixing `uglify` failures (I think poor error messages are particularly frustrating) but don't think adding an extra flag is the way to go

lordvlad:
Yeah, I'd love both to be executed on production code, but for debuggin `yo-yoify` is kind of an hinderance because the sourcemapping does not match the execution steps in all the code right around the template strings.

yoshuawuyts:
Heya, thanks for the patch - we've removed the uglify switch in the latest release tho. Closing for now!

PR #168, opened 2017-04-23
super optimize

yoshuawuyts:
Makes it so we run extra optimizations after browserify has run its course. This patch preps us to possibly include per-route rendering if we detect it's a choo app:
- __uglifyjs__ in addition to `uglifyify` which on our hello world example shaves another 5-10%
- __minify-html__ which on long docs shaves bytes and does stuff like reordering classnames so compression will handle them better
- __purify-css__ which takes the full JS bundle and strips all CSS that's not used

cc/ @toddself can you verify this doesn't break your setup? I'm not sure about the extra uglifyjs call - should we make it configurable?

### screenshot
Note that these numbers are on a "hello world" style app, if you have a lot more HTML (eg. server rendering) then the numbers shift to the other side

screen shot 2017-04-23 at 17 07 52


toddself:
I feel like the uglify should be optional at the end -- using babel to target es5 solely to satisfy uglify when your target deployment environment supports your syntax feels like a very high tax to have to pay to use bankai for the community writ large.

yoshuawuyts:
Ok, fixed! - think I also caught a bug with uglify accidentally being triggered even though `false` was passed. Check out the difference between the 3rd and 2nd render (uglify off for `bin.js` vs uglify off for `bin.js` + `index.js`)

screen shot 2017-04-23 at 17 34 18

yoshuawuyts:
`v7.5.0` :tada:

Issue #169, opened 2017-04-24
inline critical CSS


yoshuawuyts:
It'd be fun if we could inline critical CSS into each HTML document we create. I've taken a good look (read: gone through the source) of a fair number of packages that do this, and it seems the right approach is to use phantomJS to figure out what's critical.

I'd love to use an existing package, but unfortunately none live up to the standards I'm holding for this repo (e.g. doing async IO). So we'll have to write our own.

For whoever's interested in doing this, I can recommend reading the source in https://github.com/pocketjoso/penthouse/blob/master/lib/index.js - most packages indirectly depend on Penthouse, and in turn the core of what they do is basically instrumenting PhantomJS.

I think with a few hours of adapting we should be there. The API I'd like to see for this would be:
```js
var critical = require('<our package>')
var fs = require('fs')

var html = fs.readFileSync(__dirname + 'index.html')
var css = fs.readFileSync(__dirname + 'bundle.css')

// with opts
var opts = { height: 900, width: 1300 } // default viewport arguments
critical(html, css, opts, function (err, criticalCss) {
  console.log(criticalCss)
})

// without opts
critical(html, css, function (err, criticalCss) {
  console.log(criticalCss)
})
```

That's all we need. From there we can do the inlining ourselves using `hyperstream`. No need to extract from the existing CSS, I'm not sure how good of an idea it is in the first place, but it's definitely a different concern. Also no need to minify, there's minifiers for that purpose.

I hope this makes sense!

### See Also
- https://github.com/pagespeed/ngx_pagespeed
- https://github.com/pocketjoso/penthouse

yoshuawuyts:
Though in all fairness this is a non-critical feature. We can easily outsource this to `ngx_pagespeed` at no cost - also there's plenty of knobs to turn there. That said: just having this work in here would prevent the need from running complex setups at the L7 layer. And the more we can inspect our static assets before they go live, the better.

PR #170, opened 2017-04-24
server render

yoshuawuyts:
If you use Bankai with Choo this'll mean we'll render out each path on the server. I'm thinking we should probably have a way to turn this off though (flag name suggestions, anyone?)

For it to work with Choo, we need access to the router so hence https://github.com/yoshuawuyts/choo/pull/482

I was thinking of also inlining CSS #169 but yeah nah, not now. Also just doing choo rendering for now, because that's why I use - PRs for supporting other frameworks are welcome!

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -25,6 +25,7 @@ var zlib = require('zlib')
 var fs = require('fs')
 
 var htmlMinifyStream = require('./lib/html-minify-stream')
+var detectRouter = require('./lib/detect-router')
 var zlibMaybe = require('./lib/gzip-maybe')
 var Sse = require('./lib/sse')
 var bankai = require('./')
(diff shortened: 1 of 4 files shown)


s3ththompson:
I'm working on an Choo app with a massive Firebase auth dependency that's only needed for a few routes:

screen shot 2017-04-24 at 4 47 20 pm

It would be suuuuuper cool if Bankai could do automatic [route-based chunking](https://gist.github.com/addyosmani/44678d476b8843fd981ff8011d389724)... this PR seems like it might set the stage for this type of feature really well!

yoshuawuyts:
Yeah, that's one option - though to be fair I'm not sold on the whole chunking thing - getting it right is non-trivial, and the solutions I've seen so far have been dauntingly complex. Perhaps we'll start doing this eventually though (:

PR #178, opened 2017-05-08
Allow create-html options to be passed in from cli

toddself:


toddself:
So you can't actually pass in arguments to create-html from the command line since `opts.html = {_: [ 'your value']}`

This allows you to actually pass in options so you don't have to recreate what's in bin.js to add a `<meta>` tag to your file :)

lrlna:
yes plssss ✨

yoshuawuyts:
If I can be cheeky: I think this would be more consistent if it used the [subarg](https://www.npmjs.com/package/subarg) API instead:

```txt
$ bankai example.js --html [ --lang en ]
```

Think inline JSON is slightly odd haha. Thanks!

toddself:
Ooh good point. Let me switch to that. I dislike json on the cli too but was just like lets get 'er done. Lol

toddself:
Um, this is now just a doc update to illustrate how you'd pass in via subargs.

toddself:
@yoshuawuyts should we merge this in?

PR #196, opened 2017-06-02
Use local index.html for start and build commands

### LINKED ISSUE #192: [question] Allow disables in build command
Sometimes is cleaner to have a `index.html` file instead of make it through `--html` option. So I was wondering, does it make sense to allow html disable in build command, I was thinking like

```bash
$ bankai build --html=false index.js
```

that would output only `bundle.js` and `bundle.css`. Right now, the API allow to disable html and css, so it sounds logical to me to do the same for the cli. What do you think?

YerkoPalma:
fix #192 

Some notes:

- If local file is present and also html options are present, this PR ignore the whole `--html` options
    ```bash
    bankai start --html [ --title "ignored" ]
    ```
    **question**: Should this log some warning? or should the `--html` options have higher priority?
- The example now only covers one use of bankai, and there are others (like the whole api) uncovered.
    **question**: It is a good idea to have an `example` folder with all the possible uses of bankai?

That's it. Waiting for reviews :)


YerkoPalma:
@yoshuawuyts removed lock file :+1:

yoshuawuyts:
lol, merge conflict - I'm sorry haha, didn't realize :P

YerkoPalma:
Conflicts resolved

yoshuawuyts:
📦 `v8.1.0`

PR #197, opened 2017-06-12
update bankai to use uglify-es

lrlna:
This PR brings in the new-and-improved version of [uglify](https://github.com/mishoo/UglifyJS2/tree/harmony) that is meant to handle es6 syntax better than the previous versions. This merge is dependant on [uglifyify pr](https://github.com/hughsk/uglifyify/pull/65) getting merged first to avoid weirdness with the option changes (there have been a lot between the versions).

Since there have been quite a few option changes, I think we might be able to compress things a lot further with some of these. I haven't added anything aside from the options you've previously had, but if something else [out of the list](https://github.com/mishoo/UglifyJS2/tree/harmony#minify-options) peaks your interest, let me know and I can add it. 

I've also added `uglify` option to the boolean list, which might help with some of the non-boolean values you've been getting.

Thanks a bunch ✨

(cc @toddself @yoshuawuyts)

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -9,7 +9,7 @@ var concat = require('concat-stream')
 var mapLimit = require('map-limit')
 var purify = require('purify-css')
 var logHttp = require('log-http')
-var uglify = require('uglify-js')
+var uglify = require('uglify-es')
 var resolve = require('resolve')
 var mkdirp = require('mkdirp')
 var subarg = require('subarg')
@@ -40,8 +40,7 @@ var argv = subarg(process.argv.slice(2), {
     assets: 'assets',
     debug: false,
     open: false,
-    port: 8080,
-    uglify: true
+    port: 8080
   },
   alias: {
     address: 'A',
@@ -52,7 +51,6 @@ var argv = subarg(process.argv.slice(2), {
     help: 'h',
     html: 'H',
     js: 'j',
-    uglify: 'u',
     open: 'o',
     port: 'p',
     verbose: 'V',
@@ -87,7 +85,6 @@ var usage = `
       -p, --port=<n>          Bind bankai to a port [default: 8080]
       -V, --verbose           Include debug messages
       -w, --watch <bool>      Toggle watch mode
-      -u, --uglify <bool>     Toggle uglifyify. [default: true]
 
   Examples:
     $ bankai index.js -p 8080            # start bankai on port 8080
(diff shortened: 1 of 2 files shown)


toddself:
I want to verify that uglify-es doesn't mangle UTF-8 characters (emoji mostly!) so let me whip something up real quick using choo-log :)

bin.js
```diff
@@ -271,11 +271,9 @@ function build (entry, outputDir, argv, done) {
     if (argv.uglify !== false && argv.uglify !== 'false') {
       log.debug('uglify starting')
       js = uglify.minify(js, {
-        fromString: true,
-        compress: true,
-        mangle: true,
-        filename: file,
-        sourceMaps: false
+        sourceMap: {
+          filename: file
```
yoshuawuyts: I'm not sure we always want source maps enabled; we should enable stuff like `unsafe` and mangle attributes tho :sparkles:

yoshuawuyts:
```sh
❯ cat dist/bundle.min.js | uglifyjs --mangle-props -c unsafe,
properties,dead_code,comparisons,evaluate,hoist_funs,if_return,
join_vars,pure_getters,reduce_vars,collapse_vars --toplevel | less
```

yoshuawuyts: probably enable source maps only if the `--debug` flag was passed

lrlna: Unless I missed something, I am not sure if there is another way to pass in the filename except through `sourceMap`, which is why I kept it there. 

that `cat`, are those the opts you're thinking of keeping enabled?

yoshuawuyts: yeah those are the options, and do we need the filename? - just gotta make sure we're not including source maps unless through `--debug` haha :sparkles:

lrlna: i guess then we are looking at two different set of options getting passed on based on whether it's `--debug`, yea?

lrlna:
@yoshuawuyts ok, added extra opts and adding sourceMaps only if in `--debug` :sparkles:

lrlna:
ok, lemme test one more time on a couple local repos to be sure, and then do you want to merge?

yoshuawuyts:
@lrlna ya 👌

lrlna:
@toddself we r g00d with the UTF-8 chars:
screen shot 2017-06-14 at 12 46 37

@yoshuawuyts ok, tested on a few more repos, seems to have no glitches. Also added `{ global: true }` to uglifyify opts just like we did in the [uglifyify PR](https://github.com/hughsk/uglifyify/pull/65) and upped its version to `4.0.0` since it's been already merged. I think we should be good to go ?

yoshuawuyts:
Oh, given that this is probably a semver major - we should probs remove the `--uglify` switch from `bankai`. No reason not to have it enabled now haha :sparkles:

toddself:
Yeah, lets drop that switch as it was solely to avoid babel when you don't want it!

lrlna:
@toddself yea! it's done :sparkles:

yoshuawuyts:
`v8.0.0` :sparkles:

lrlna:
✨✌️✨✌️✨

PR #199, opened 2017-06-19
fix uglify

yoshuawuyts:
couldn't get my project to build, this fixes it
- only run uglifyify
- throw err on `--verbose`
- more debug statements
- remove broken flag

cc/ @lrlna


lrlna:
yea there is an issue with [uglifyify](https://github.com/hughsk/uglifyify/issues/66) that i've been debugging for the past few days; doesn't seem like your bug was related tho, eh?

PR #201, opened 2017-06-19
Use relative path for build command. fix #198

### LINKED ISSUE #198: Wrong path resolution on building
I have a project structured like this 

```text
project/
├── api/
├── public/
├── src/
└── server.js
```

My `api` folder has my backend, and my `src` folder my frontend source, in my `public` folder is my _compiled_ frontend. I can serve my frontend stuff with `bankai start`, but I can't use it to build my stuff. If I run:

```bash
$ bankai build --assets=assets src/index.js public/
```

I get an error saying that the folder `src/public/assets` is not found. If I run 

```bash
$ bankai build --assets=assets src/index.js ../public/
```

I get my assets build in my public folder, and a new public folder _outside_ of my root project folder with my html, js and css files.

What is happening, is that for building html, js and css, bankai is taking as root the current folder, but when copying assets, it uses as root the entry file folder. If your entry point is in your root directory this would never be an issue, if you use a structure like the one I show (or like what express or other full stack frameworks do) it will have this behaviour.

YerkoPalma:


YerkoPalma:
@yoshuawuyts are you ok with this change?

yoshuawuyts:
Oh no, sorry for not reviewing sooner - patch LGTM, but errr - we just rewrote the whole codebase haha. Lmk if the v9 beta works out alright for you. Thanks!

PR #202, opened 2017-06-20
Readme update

emkay:
Since the uglify option was removed I think it makes sense to take it out of the docs.


yoshuawuyts:
📦 `v8.1.1`

PR #203, opened 2017-06-23
don't uglify in debug mode

lrlna:
It's easier to track down values and variables in debug / development mode when uglify is turned off. This PR doesn't uglify in debug.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,8 +104,7 @@ function Bankai (entry, opts) {
         collapse_vars: true
       }
     }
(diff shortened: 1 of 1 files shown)


toddself:
it would be nice to be able to turn off uglify (or better control mangling?) -- having code that you're trying to debug optimized away is mildly frustrating.

yoshuawuyts:
Perhaps a generic flag to turn off transforms might be the way to go?
Transforms are great, but you're right in that software isn't flawless, and
having a way to toggle things might be useful

PR #206, opened 2017-06-26
HTTPS is the way

toddself:
That `.default` thing caught me for a minute.  Really wish there was better
interop between cjs and ecmascript modules... Not sure if the `.default` is 
better on the require (so the API works as "expected" w/r/t to cjs) or if you
want `devCert.default()` when invoking it.

Plus the `{installCertutil: true}` option doesn't seem to work on my mac at the
very least, as [chrome and firefox still give security warnings](https://github.com/davewasmer/devcert/issues/4)?

Lastly should we be only HTTPS or HTTP by default with HTTPS opt-in or HTTPS
by default or HTTP opt-in

I figure if we can get this working on multiple platforms without security
issues there's no reason not to do HTTPS-only?

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -9,13 +9,14 @@ var concat = require('concat-stream')
 var mapLimit = require('map-limit')
 var purify = require('purify-css')
 var logHttp = require('log-http')
+var devCert = require('devcert').default
 var resolve = require('resolve')
 var mkdirp = require('mkdirp')
 var subarg = require('subarg')
 var tmp = require('temp-path')
+var https = require('https')
 var from = require('from2')
 var disc = require('disc')
-var http = require('http')
 var open = require('open')
 var path = require('path')
 var pino = require('pino')
@@ -146,6 +147,10 @@ function start (entry, argv, done) {
   var address = argv.address
   var port = argv.port
   var sse = Sse(assets)
+  var certOpts = {
+    installCertutil: true
+  }
+  var certName = entry.replace(new RegExp(path.sep, 'g'), '-').slice(-64)
 
   assets.on('js-bundle', function () {
     log.info('bundle:js')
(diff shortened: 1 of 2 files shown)


bin.js
```diff
@@ -146,6 +147,10 @@ function start (entry, argv, done) {
   var address = argv.address
   var port = argv.port
   var sse = Sse(assets)
+  var certOpts = {
+    installCertutil: true
+  }
+  var certName = entry.replace(new RegExp(path.sep, 'g'), '-').slice(-64)
```
toddself: The certname should be unique for each project (maybe per entry point?) -- but there's a 64 character limit on it due to ASN1:

```
↳ openssl req -config /Users/todd/.config/devcert/openssl.conf -subj "/CN=-Users-todd-src-showrunner-node_modules-@scriptollc-code-mirror-ot-standalone.js" -key /Users/todd/.config/devcert/-Users-todd-src-showrunner-node_modules-@scriptollc-code-mirror-ot-standalone.js.key -out /Users/todd/.config/devcert/-Users-todd-src-showrunner-node_modules-@scriptollc-code-mirror-ot-standalone.js.csr -new
problems making Certificate Request
4550:error:0D07A097:asn1 encoding routines:ASN1_mbstring_ncopy:string too long:/BuildRoot/Library/Caches/com.apple.xbs/Sources/OpenSSL098/OpenSSL098-64.50.6/src/crypto/asn1/a_mbstr.c:154:maxsize=64
```

toddself: So we have to verify that we only ever pass in 64 characters.  If you try to pass in with path separators, it fails as well since the cert name has the directory name attached, and that directory doesn't exist (which might be a bug in devcert).

toddself:
We might also want to wait for:
https://github.com/davewasmer/devcert/pull/7
https://github.com/davewasmer/devcert/pull/8

To land (if they do) since we can stop doing the data manipulation and let it occur upstream (which feels more proper).

yoshuawuyts:
Ohey, yeah alternatively we could just use https://github.com/Dexus/pem - it's what `wzrd` uses and means we don't have to deal with weird source code

toddself:
Closed in favor of #217

PR #213, opened 2017-07-09
Inline critical css

yoshuawuyts:
Inlines critical CSS. Closes https://github.com/choojs/bankai/issues/169. Not super optimal, but hey it's close enough (and doesn't require headless chrome/ electron or similar). Also adds a `<title>` element if `state.title` is set in choo.

Relies on choo 6, and https://github.com/sethvincent/create-html/pull/9 (although we're working off a branch now).

Result is that we're now 100/100 on 2/4 marks. HTTP/2 and service workers should get us too 100/100 on all 4. Whoop! ([link to viewer](https://googlechrome.github.io/lighthouse/viewer/?gist=c1f7a56600fbe811c3e0237f1b7719e9))

screen shot 2017-07-09 at 16 50 14


yoshuawuyts:
cc/ @choojs/trainspotters

PR #216, opened 2017-07-13
Add support for HTTPS

toddself:
Adds support to create a self-signed certificate and provide it for starting
an HTTPS server


toddself:
@choojs/trainspotters right now you have to accept the cert via the browser's cert stuff and it never gets saved to disk.

Should we provide an option that allows it to be written to disk so you could add it to a local trust store?  O an option to provide your own SSL certs?

I thought best to keep it simple for now, and then figure out advanced use cases for later.

yoshuawuyts:
Not bad! Few points I'm wondering about:
- with this approach people will always need to click past the "insecure key" warning screen
- SSL is disabled by default, adding yet another option
- HTTP isn't redirected to HTTPS

Ideally this would be enabled by default, and we'd have a secure key. Thinking a way to do this might be perhaps to solve the ACME challenge from letsencrypt, and include that. Would mean no unsafe self-signed keys; like a reasonable approach. Alternatively we could ask people to trust a root signing key, and save it to disk - but that seems, like, hairy.

Thoughts?

toddself:
So. Yeah, this is difficult.  We can't do ACME since its localhost.

We should do the SSL on by default and redirect to HTTPS from HTTP, but unless we modify the user's certificate store for them (a slippery slope of hard-to-use APIs) they're going to either have to add a trust for the key we generate locally on each machine or click through the security warning.

I think some instructions and letting people choose to trust the key we generate (and store locally for future requests) or have them click through are the best options?

yoshuawuyts:
Think having them trust a key we generate for them is the right approach.
Include instructions on how to do it in a wiki of sorts. (:

yoshuawuyts:
Think we can safely close this as HTTPS has landed as part of the Bankai v9 beta. :tada: Thanks @toddself :D

PR #231, opened 2017-08-26
Allow passing state through to html

justinj:
This allows one to pass any state they'd like included into their SSR.

I'm using this to affect the pre-rendering, along with dumping all the state into a hidden input to actually grab it from the client. Not sure if there's a preferred way!

### DIFF
--- a/index.js
+++ b/index.js
@@ -151,7 +151,7 @@ Bankai.prototype.js = function (req, res) {
 }
 
 // (obj, obj) -> readStream
-Bankai.prototype.html = function (req, res) {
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Hey! We're currently doing a full rewrite in the https://github.com/choojs/bankai/tree/new-new-new branch. Could you perhaps file a patch against there? Thanks! :grin:

justinj:
Ok! Since it seems like that's not quite ready for general use yet, I think I'll hold off until it is :)

PR #241, opened 2017-09-20
Bankai rewrite

yoshuawuyts:
Everything is new! Would be cool if people could take this for a test ride so we can fix bugs. Thanks!

## Bankai start
bankai2

## Bankai build
bankai3

## Known issues
- [x] `SIGTERM` can leave the cursor as hidden
- [x] extracting critical CSS does only detects classes, not tag name selectors
- [x] reloading scripts for SSR in `document` can crash, causing stale builds
- [x] SIPHASH24 sometimes generates a new hash for the same files (could be our browserify pipeline too, need to investigate)


lib/http-server.js
```diff
@@ -0,0 +1,154 @@
...
+var mkdirp = require('mkdirp')
+var http = require('http')
+var path = require('path')
+var pump = require('pump')
+var net = require('net')
+var fs = require('fs')
+var os = require('os')
+
+var selfsigned = require('selfsigned')
+var getPort = require('getPort')
```
goto-bus-stop:
```
$ node ../../choojs/bankai/bin.js app.js
module.js:529
    throw err;
    ^

Error: Cannot find module 'getPort'
```

This require should be all lowercase `getport`

YerkoPalma: Found that too, there is a typo there `getPort` -> `getport`

YerkoPalma:
Seems like tests are not passing for node 4

lib/reload-client.js
```diff
@@ -0,0 +1,61 @@
...
+      return log.error('error parsing event', e)
+    }
+    if (ev.type === 'style') handleStyles()
+    else if (ev.type === 'scripts') handleScripts()
+  }, false)
+  timing()
+
+  function handleScripts () {
+    log.info('scripts', 'reloading')
+    window.location.reload()
```
bendik: Reloading causes a SSE connection broken error in console, at least in Firefox.
How about adding this?
```js
window.addEventListener('beforeunload', function (event) {
  source.close()
});
```

yoshuawuyts: Good call, fixed! :D

lib/http-server.js
yoshuawuyts: Good catch; thanks!

yoshuawuyts:
Looks like this is caused by `npm` rather than Node versions. It seems that for whatever reason the `documents:list` method never fires; not sure why.

ajoslin:
With this and goto-bus-stop's work, Browserify is becoming OP again :) 

Big thanks Yosh, I can't wait to try this out.

thlorenz:
I may have the solution for your cursor problem. Wrote [this](https://github.com/thlorenz/cursor-back) a while back when I ran into similar problems. Not sure if it'll handle `SIGTERM` though (it has been a while).

yoshuawuyts:
@thlorenz thanks for the pointer, fixed the cursor issue :D

jbergstroem:
FWIW this doesn't work great with `pnpm` (`npm` replacement). Likely due to a missing dep and npm's flattening assumptions. I can't look into this right now -- expect tomorrow.

```console
~/oss/bankai $ ./bin.js build -d example
14:30:17 ✨  created: ./dist
14:30:17 🚨  bankai.asset: could not find a file for 
14:30:17 ✨  created: ./dist/manifest.json
 
~/oss/bankai ±pr/241 $ echo $?
0
```

yoshuawuyts:
@jbergstroem !!! that's great news; I thought it might be a missing file, but `dependency-check` didn't return any positives. Given this is true, it would fix it for `node@4` too! :D

YerkoPalma:
There are some options that now are not supported by bankai. Are they completly erased now? for instance the `port` and `address` options? Also there are differences from the api and the cli, for instance, how could you manipulate `assets` path from the cli? it seems that you can't but it can be done with the node api

jbergstroem:
This is also a bit concerning imo (slow):
```console
$ time node bin.js   
  Please specify a bankai command:
    $ bankai <command>

  For example:
    $ bankai start index.js

  Run bankai --help to see all options.
node bin.js  0.84s user 0.11s system 101% cpu 0.940 total

$ time npm help   
<snip>
npm help  0.19s user 0.07s system 76% cpu 0.337 total
```

yoshuawuyts:
@yerkopalma Ports are now automatically generated (`8080-9000`). The assets path has become more of a convention thing - picking up `assets/` and `content/`. We haven't set extra paths yet for the JS API, but I think that'd be a good addition - although if you have some common paths in mind we could add those to the CLI too

@jbergstroem made the imports in `bin.js` dynamic; should be sped up by a whole bunch!

jbergstroem:
I seem to be getting different hashing (clean repo). MacOS 10.13 17A362a, node 8.5.0:
```console
not ok 34 ·· ·· <script src="/scripts/c4bc18653915ae8c/bundle.js" defer>
  ---
    operator: equal
    expected: '<script src="/scripts/c4bc18653915ae8c/bundle.js" defer>'
    actual:   '<script src="/scripts/75023dc608219a4e/bundle.js" defer>'
    at: assertHtml (/Users/jbergstroem/wrk/oss/bankai/node_modules/assert-html/index.js:82:15)
```

yoshuawuyts:
@jbergstroem bizarre; that could only be if for some reason WASM is acting up, the input files have changed, or like the algo isn't working. Dep for that is in https://github.com/yoshuawuyts/buffer-graph/ and https://github.com/mafintosh/siphash24 - we might need to switch to blake2b tbh; will probably be more stable

jbergstroem:
@yoshuawuyts you probs want this:
```diff
--- test/service-worker.js      2017-09-20 18:39:16.000000000 -0300
+++ test/service-worker-new.js  2017-09-20 18:39:11.000000000 -0300
@@ -47,7 +47,7 @@
     assert.ok(res.hash, 'output hash exists')
   })
 
-  compiler.script('bundle.js', function (err, res) {
+  compiler.scripts('bundle.js', function (err, res) {
     assert.error(err, 'no error writing script')
     rimraf.sync(tmpDirname)
   })
@@ -72,7 +72,7 @@
     assert.ok(res, 'output exists')
   })
 
-  compiler.script('bundle.js', function (err, res) {
+  compiler.scripts('bundle.js', function (err, res) {
     assert.error(err, 'no error writing script')
     rimraf.sync(tmpDirname)
   })
```

..and add it to `index.js`?

jbergstroem:
> @yoshuawuyts said:
>  it definitely sounds as if the files have somehow changed

At least the repo is intact:
```console
$ rm -rf node_modules && npm i -q && git reset --hard && node test/document.js 

> fsevents@1.1.2 install /Users/jbergstroem/wrk/oss/bankai/node_modules/fsevents
> node install

[fsevents] Success: "/Users/jbergstroem/wrk/oss/bankai/node_modules/fsevents/lib/binding/Release/node-v57-darwin-x64/fse.node" already installed
Pass --update-binary to reinstall or --build-from-source to recompile
added 885 packages in 46.893s
HEAD is now at e6f1152 make it fast
<snip>
not ok 31 ·· ·· <script src="/scripts/c4bc18653915ae8c/bundle.js" defer>
  ---
    operator: equal
    expected: '<script src="/scripts/c4bc18653915ae8c/bundle.js" defer>'
    actual:   '<script src="/scripts/75023dc608219a4e/bundle.js" defer>'
    at: assertHtml (/Users/jbergstroem/wrk/oss/bankai/node_modules/assert-html/index.js:82:15)
    stack: |-
      Error: ·· ·· <script src="/scripts/c4bc18653915ae8c/bundle.js" defer>
          at Test.assert [as _assert] (/Users/jbergstroem/wrk/oss/bankai/node_modules/tape/lib/test.js:212:54)
          at Test.bound [as _assert] (/Users/jbergstroem/wrk/oss/bankai/node_modules/tape/lib/test.js:64:32)
          at Test.equal.Test.equals.Test.isEqual.Test.is.Test.strictEqual.Test.strictEquals (/Users/jbergstroem/wrk/oss/bankai/node_modules/tape/lib/test.js:347:10)
          at Test.bound [as equal] (/Users/jbergstroem/wrk/oss/bankai/node_modules/tape/lib/test.js:64:32)
          at assertHtml (/Users/jbergstroem/wrk/oss/bankai/node_modules/assert-html/index.js:82:15)
          at /Users/jbergstroem/wrk/oss/bankai/test/document.js:99:5
          at /Users/jbergstroem/wrk/oss/bankai/index.js:127:5
          at /Users/jbergstroem/wrk/oss/bankai/lib/queue.js:24:5
          at Array.forEach (<anonymous>)
          at Queue.ready (/Users/jbergstroem/wrk/oss/bankai/lib/queue.js:23:13)
  ...

**Comment by s3ththompson** [comment]
This looks amazing! So excited to swap it out in my current projects.

Would be nice if there was a quick JS API migration guide... should I just copy over the contents of `cmd-start`? Do you plan to leave handling gzip and content-type headers at the userland level?

**Comment by yoshuawuyts** [comment]
@s3ththompson oh boy; yeah so the whole API was rewritten from scratch. I'm not sure how much of a migration guide could exist, I'd mostly just suggest to read the docs.

That said, we should probably document the different directories we look for; same for files. Probably also document some of the flags that aren't documented yet (`watch`, `reload`), and errr - yeah that's about it? On it! :grin:

**Comment by shotlom** [comment]
Can I suggest some little attention to  #242 & #243 before pulling. Sorry. But many ★★★★★ for the direction.

**Comment by yoshuawuyts** [comment]
@shotlom fixed!

**Comment by YerkoPalma** [comment]
Why is there a `'other'` file when building?

**Comment by ajoslin** [comment]
JS compile errors aren't showing up:

index.js:
```
var errorVar = '1
```

Both `bankai start` and `bankai build` fail silently without showing the compile error.

`bankai` start doesn't show the error whether I pass in `--debug` or not.

**Comment by yoshuawuyts** [comment]
@ajoslin ah yeah, great point. I forgot to restore the code for that. I'll add the code back for `bankai build`; adding it for `bankai start` might be a bit trickier, because to do it right we'd need to add a separate view buffer (e.g. wanna do a multi-tab window layout). It might be easier to restore with the `-q` flag tho, because it means we can skip that mode. Going to fix this!

**Comment by yoshuawuyts** [comment]
@YerkoPalma that's an SSR HTML file. Besides `index.html` the files can't really have an `.html` extension - so this means that going to `localhost:8080/other` just works when served from disk :sparkles:

**Comment by s3ththompson** [comment]
@yoshuawuyts makes sense that a migration guide might not be applicable, but I'm still trying to figure out what stuff bankai no longer does by default and what stuff I should do at userland level...

As far as I can tell all transforms that were previously built into both JS API and CLI (e.g. yoyoify, uglify, and envify) are now only in the CLI?

My use case is wanting to create a server for my app that behaves identically to the CLI, but also exposes its own routes and endpoints...

**Comment by s3ththompson** [comment]
@yoshuawuyts or to put it differently, it seems like there's now a significant difference between the behavior/functionality of the API and CLI.

**Comment by yoshuawuyts** [comment]
> As far as I can tell all transforms that were previously built into both JS API and CLI (e.g. yoyoify, uglify, and envify) are now only in the CLI?

Actually the inverse is true! - In the past we only had certain transforms in the CLI, but not in the JS API. This rewrite was done so we could clean up all the custom glue code with a graph structure, and make sure that all transforms run everywhere! :sparkles:

**Comment by s3ththompson** [comment]
Ahh, was just reading the code, and didn't see that `lib/graph-script.js` was imported by both CLI & API. nvm, makes sense now.

So it's just [`gzip-maybe` ](https://github.com/choojs/bankai/blob/new-new-new/lib/cmd-start.js#L224) that I'll have to add manually if I want identical API / CLI behavior?

**Comment by yoshuawuyts** [comment]
@YerkoPalma it now skips over all dynamic routes, as Bankai can't possibly figure out what the right pathname is for that. For dynamic routes SSR people should run bankai dynamically I think.

@s3ththompson oh actually, good point! - I was meaning to add `.gz` and `.deflate` versions of all non-asset files. That way servers like Nginx / Caddy can pick them up straight away when deployed :sparkles:

**Comment by jbergstroem** [comment]
@yoshuawuyts: `nanologger` is the missing dependency. Add it to `package.json` and pnpm is at least happy. Perhaps @YerkoPalma can test on node 4 as well.

Also, 🏅 if you support `.br` as well :)

**Comment by jbergstroem** [comment]
@yoshuawuyts on another note; I get a _third_ hash now 😢 
```console
    expected: '<script src="/c4bc18653915ae8c/bundle.js" defer>'
    actual:   '<script src="/770e20fd242c6281/bundle.js" defer>'
```

..I'll have a look at whats going on, but there should be no webassembly shenanigans.

YerkoPalma:
Mmm not sure if I'm right but this looks like prerendering to me, more than SSR, since SSR is dynamic.
However, this means that by default `bankai build` will not build dynamic routes? (at least I couldn't do it)

@jbergstroem I just ran `npm t` with node 4 and all test pass

yoshuawuyts:
@s3ththompson `bankai build` now compresses too! :tada:

@jbergstroem wooooooot, fixed! :tada: Also got `.br` going on :sparkles: :grin:

yoshuawuyts:
> However, this means that by default bankai build will not build dynamic routes? (at least I couldn't do it)

@YerkoPalma Yeah, that's correct. If you have a suggestion on how to approach this, I'd be open. But all that I've come up with definitely exceeds what a static build tool can and probably should do.

yoshuawuyts:
@jbergstroem yeah, got a hunch that it might actually be that some browserify transform is non-deterministic, which causes the output to change. I'm not sure which one tho, but ya :/

A way to detect it would probably be to run `diff(1)` on both files, and check what has changed between the two. If there's no changes, it means that it isn't browserify, but definitely the hashing algo.

jbergstroem:
@yoshuawuyts disabling web assembly doesn't help at least (`--no-expose_wasm`). I'll dig in.

s3ththompson:
@YerkoPalma @yoshuawuyts i've been thinking of the dynamic routes too... I think one simple solution would be an option to pass a list of routes to bankai that it attempts to prerender and save. so if I have a route `/products/:id`, I could say `/products/32` and `/products/88` are my most popular urls, so i'd like to have bankai prerender those...

YerkoPalma:
> But all that I've come up with definitely exceeds what a static build tool can and probably should do.

But previous versions of bankai handle dynamic routes without a problem, and is a very common patterns in SPAs, what's different here is the prerendering which I think is definitely out of scope in dynamic routes for a static build app like bankai.

Not sure right now what would be a good solution :/

yoshuawuyts:
I wrote https://github.com/yoshuawuyts/extract-html-tag, and updated https://github.com/yoshuawuyts/inline-critical-css to use it, which means CSS inlining is fixed! :tada: Just need to fix the `decache` bug, figure out the hashing problem and I think we should be good for a beta release :tada:

jbergstroem:
> @yoshuawuyts said:
> figure out the hashing problem

I might have some time for this today.

jbergstroem:
Just did a quick test on both freebsd and in alpine linux. In FreeBSD (node 8.x) I get a different hash. Alpine (node 6) correctly passes the test.

yoshuawuyts:
Thats heaps interesting! Could you perhaps get the build output of both
into a file? It'd be good to compare and figure out if the output is
different.

lib/graph-manifest.js
```diff
@@ -0,0 +1,60 @@
...
+        createEdge('color', DEFAULT_COLOR)
+        createEdge('bundle', DEFAULT_MANIFEST)
+        return
+      }
+
+      fs.readFile(filename, function (err, file) {
+        if (err) return self.emit('error', 'manifest', 'fs.readfile', err)
+
+        var res = parseJson(file)
+        if (err) return self.emit('error', 'manifest', 'JSON.parse', err)
```
dberesford: should this be `if(res.err)`

jbergstroem:
> @yoshuawuyts said:
> Thats heaps interesting! Could you perhaps get the build output of both into a file? It'd be good to compare and figure out if the output is different.

I did just that. SHAsumming the files shows that the content (expected vs generated) is identical.

s3ththompson:
looks like it's missing `mkdirp` as a dependency in `package.json`...

bendik:
Noticed that Safari doesn't seem to be to happy with the rel="preload" onload="this.rel='stylesheet'" tactic for some reason

s3ththompson:
I'm unable to build my app due to this `buffer-graph`/`siphash24` error: https://github.com/yoshuawuyts/buffer-graph/issues/2

lib/graph-manifest.js
yoshuawuyts: Good catch!

yoshuawuyts:
@bendik Oh yeah, good catch; we could actually use help with this! - Ideally the FT CDN would cover this case (see https://github.com/Financial-Times/polyfill-service/issues/981). Until then we should probably always serve up a shim for this specific case. Not too thrilled about it, but this should definitely work on Safari.

lib/graph-manifest.js
yoshuawuyts: Nice catch!

yoshuawuyts:
>  perhaps also worth mentioning: the precompiled iltorb doesn't really play on alpine linux/musl (which is a pretty common choice for building inside containers)

Damn, that's not great. Any suggestions on what we could do to solve this? Perhaps raise in the `iltorb` module? Feature detect?

yoshuawuyts:
> Edit: the generated index.js seems identical (shasum match) across all os'es.

@jbergstroem did you mean `bundle.js`? Thanks for digging into this, this is heaps helpful! Talked to Mathias and raised in the `siphash24` repo so we can have an isolated discussion about it there. There's also https://github.com/mafintosh/siphash24/issues/2 which is under investigation.

jbergstroem:
> @yoshuawuyts said:
> @jbergstroem did you mean bundle.js?

I was referring to the file generated as part of the test suite.

> @yoshuawuyts said:
> Damn, that's not great. Any suggestions on what we could do to solve this? Perhaps raise in the iltorb module? Feature detect?

This is pretty much consistent for all packages using gyp/pre-gyp 😞 -- safest way is [forcing a rebuild](https://github.com/mapbox/node-pre-gyp#options) (`--build-from-source`), but that penalizes most other use-cases by adding more time at installation. The other trade-off is going back to a pure-js implementation. I'd be ok with that seeing how rarely compression should be invoked (production environments).

jbergstroem:
Just did a simple hash replacement test and ran into a similar issue (hashes generated on macOS -> test run from alpine):
```console
not ok 34 ·· ·· <script src="/c3582fadc2302300/bundle.js" defer>
  ---
    operator: equal
    expected: '<script src="/c3582fadc2302300/bundle.js" defer>'
    actual:   '<script src="/3e0cdabfde62d4b1/bundle.js" defer>'
    at: assertHtml (/bankai/node_modules/assert-html/index.js:82:15)
```

So, hashing itself is likely not the culprit.

yoshuawuyts:
Welp, source maps are working now! :sparkles:

jondashkyle:
Couple quick ones:

- Looks like I’m unable to access files in the `/assets` directory. I have a few images in there, but when navigating to `/assets/image.jpg` I see the error `bankai.asset: could not find a file for assets/image.jpg`.

- Not so much a bug as much as it is a possible improvement, but wondering if there might be a way to speed up the reload functionality. Compared to something like budo, there is quite a lag.

lib/graph-script.js
```diff
@@ -46,6 +47,7 @@ function node (state, createEdge) {
   b.transform(yoyoify, { global: true })
   b.transform(envify, { global: true })
   b.transform(uglifyify, uglifyOpts(entry))
+  b.plugin(commonShake)
```
jbergstroem: Shouldn't this be called pre-uglify? If I recall correctly, Indutny's tree shaker only comments code.

goto-bus-stop: oh yeah that's a good catch, it needs to be uglified afterward--uglifyify runs as a transform though, which always happens _before_ the plugin runs. so the final bundle should be passed thru uglify in order to actually clean the dead code bits.

yoshuawuyts: Fixed!

yoshuawuyts:
@jondashkyle thanks for pointing these out!

---

- __entry file:__ I can't think of a way around this at the moment. Even if we'd allow you to pass a `--basedir` flag to set a basedir, we'd run into trouble. At the moment the only way to approach this I can recommend would be to move `index.js` to the root of your project. Sorry :(
- __large files:__ yeah, was worried this might happen. Node only has, like 1.5GB of memory - anything beyond that would make it crash. You might want to try again, because we just moved off the WASM build (which might be what caused your crash to happen) and onto a regular JS one. The longer term solution for this will probably be to never copy the asset files into memory tho (especially with stuff like video, haha).
- __live reload:__ yeah, noticed that too. I think after this patch lands, and we get some of the other bugs fixed we should do a pass on performance. I'm pretty sure we can speed this up!

Hope this makes sense; thanks again for chiming in! :sparkles:

yoshuawuyts:
Hmmmmm, we definitely seem to be having some sort of race condition in our test somewhere. Something is throwing off the hash of `bundle.js`- but only sometimes

http.js
```diff
@@ -0,0 +1,254 @@
...
+      })
+      res.write(`id:${id++}\ndata:${msg}\n\n`)
+    }
+  })
+
+  router.default(function (req, res, next) {
+    var url = req.url
+    compiler.documents(url, function (err, node) {
+      if (err) {
+        return compiler.documents('/404', function (err, node) {
```
lrlna: i was just looking this over trying to implement a default: you've got two `err`, the other one should probably be `error` or something

yoshuawuyts: o/ Thanks for reporting - this is fine tho! The closures work alright in this case; there's no bugs here (:

yoshuawuyts:
Soooo, all boxes are ticked for an initial beta release? Can people please verify this works on their code?

Biggest follow ups for next patches in the beta will be:
- [ ] improve our `error` story. Right now if a crash happens, it can fail without warning which is super frustrating for people
- [ ] figure out a better story for assets, so we don't overflow memory
- [ ] perhaps sneak in stuff like `pack-flat` for browserify, if source maps get the fix
- [ ] figure out `pre-gyp` on Windows for `.br` files; this one's tricky.

Holler if I've missed anything! Thanks all :grin:

yoshuawuyts:
TESTS ARE PASSING. SSR RELOADING WORKS. The time for final reviews is now. Unless something is totally broken for someone here, we should be good to merge and put out an initial beta release! :tada:

ajoslin:
@yoshuawuyts errors are still not being outputted, both for `bankai start` and `bankai build`.

s3ththompson:
@yoshuawuyts same wasm RangeError with blake2b (will open issue). can we start with a pure JS solution and upgrade to wasm when the kinks are worked out? i think there are still some subtle issues with wasm heap size that require some investigation...

jbergstroem:
@yoshuawuyts sorry to say that hashing still seems to be a problem. I don't have too much time to dedicate to this but I'll make an attempt at debugging it soon.

yoshuawuyts:
@ajoslin you're totally right; hope this is better!

screen shot 2017-09-28 at 15 33 32

yoshuawuyts:
Okay, so I'm not thrilled about the lastest pass on errors - but getting it 100% right means implementing proper progress bar logic, more control over re-rendering, and a partial rewrite of the way we create UI (might need to ditch `ansi-diff-stream`). It's a little over the top haha.

So kinda feel that for now error handling is good enough for a beta; going to fix the blake2b error and then we should be good!

yoshuawuyts:
@s3ththompson reverted the change, range error should be fixed again!

Welp, that concludes this lil episode. Gonna merge this patch and publish as a beta release. Any feedback can be sent to https://github.com/choojs/bankai/issues/246. Thanks!

yoshuawuyts:
📦 `v9.0.0-rc1` :sparkles:

s3ththompson:
@yoshuawuyts RangeError fixed! thank you!

Issue #247, opened 2017-09-29
Error in error printing code


s3ththompson:
In trying to print this error...
```
{ Error: Cannot find module './static/css/base.css' from '/Users/seththompson/code/oy-spheres/app' while parsing file: /Users/seththompson/code/oy-spheres/app/index.js
    at Function.module.exports [as sync] (/Users/seththompson/code/oy-spheres/node_modules/resolve/lib/sync.js:33:11)
    at resolveSync (/Users/seththompson/code/oy-spheres/node_modules/style-resolve/index.js:22:19)
    at extractImportNodes (/Users/seththompson/code/oy-spheres/node_modules/sheetify/transform.js:112:27)
    at extractNodes (/Users/seththompson/code/oy-spheres/node_modules/sheetify/transform.js:85:7)
    at walk (/Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:49:9)
    at /Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:46:17
    at forEach (/Users/seththompson/code/oy-spheres/node_modules/foreach/index.js:12:16)
    at walk (/Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:34:9)
    at /Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:41:25
    at forEach (/Users/seththompson/code/oy-spheres/node_modules/foreach/index.js:12:16)
    at /Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:39:17
    at forEach (/Users/seththompson/code/oy-spheres/node_modules/foreach/index.js:12:16)
    at walk (/Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:34:9)
    at module.exports (/Users/seththompson/code/oy-spheres/node_modules/falafel/index.js:50:7)
    at DestroyableTransform.end [as _flush] (/Users/seththompson/code/oy-spheres/node_modules/sheetify/transform.js:59:13)
    at DestroyableTransform.<anonymous> (/Users/seththompson/code/oy-spheres/node_modules/readable-stream/lib/_stream_transform.js:138:49)
    at Object.onceWrapper (events.js:314:30)
    at emitNone (events.js:105:13)
    at DestroyableTransform.emit (events.js:207:7)
    at prefinish (/Users/seththompson/code/oy-spheres/node_modules/readable-stream/lib/_stream_writable.js:596:14)
[...]

PR #253, opened 2017-10-01
fix tty-err for unknown types

yoshuawuyts:
Fixes TTY error for unknown types. Closes https://github.com/choojs/bankai/issues/247. Thanks!

### DIFF
--- a/lib/tty-error.js
+++ b/lib/tty-error.js
@@ -8,6 +8,8 @@ var cwd = process.cwd()
 module.exports = ttyError
 
 function ttyError (src, sub, err) {
+  if (!err.filename || !err.loc) return err
+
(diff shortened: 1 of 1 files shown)


lib/tty-error.js
```diff
@@ -8,6 +8,8 @@ var cwd = process.cwd()
 module.exports = ttyError
 
 function ttyError (src, sub, err) {
+  if (!err.loc) return err
```
s3ththompson: I woul do `if (!err.filename || !err.loc) return err`. (See screenshot below for one of the errors I was running into...)

screen shot 2017-10-01 at 10 21 10 am

s3ththompson: otherwise LGTM

PR #256, opened 2017-10-02
Allow for <esc> and q to quit server

toddself:
Adds a listener for stdin and runs exit stuff if q or esc are pressed.


yoshuawuyts:
@toddself could you please remove `.bin.js.swp`?; perhaps add it to the `.gitignore` file. Thanks!

PR #259, opened 2017-10-03
log SSR errors, closes #255

### LINKED ISSUE #255: Server render swallows errors
This one took a little while to figure out :P

[This try/catch block](https://github.com/choojs/bankai/blob/master/lib/server-render.js#L28) effectively swallows any errors in server-side rendering your app (thus causing only the index to be rendered). Is there any reason that we don't want to print out these errors? Is `proxyrequire` noisy/flaky?

Once I manually dumped the error, it turned out that none of my routes were rendering because I was including a library that assumed the existence of `window.document`.

> Side note: It seems as though the `proxyrequire` library doesn't work with `node --inspect`... Whenever I had a DevTools window open for debugging, Bankai ran forever, stuck somewhere in the build process. This made setting breakpoints mostly impossible. Are there any `proxyrequire` alternatives?

s3ththompson:
Prints SSR errors. 

I'm not sure if we have enough information to make them prettier in `lib/tty-error.js`, so this is just a start in logging them...

### DIFF
--- a/lib/cmd-build.js
+++ b/lib/cmd-build.js
@@ -33,7 +33,9 @@ function build (entry, opts) {
 
     compiler.on('error', function (topic, sub, err) {
       if (err.pretty) log.error(err.pretty)
-      else log.error(`${topic}:${sub}`, err)
+      else {
+        log.error(`${topic}:${sub} ${err.message}\n${err.stack}`)
+      }
     })
 
     compiler.on('change', function (nodeName, edgeName, nodeState) {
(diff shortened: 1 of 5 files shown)


s3ththompson:
werkin on it... Any idea how to log unhandled error events from tape?

yoshuawuyts:
Something like this?
```js
thing.on('error', function (err) {
  assert.fail(err)
})
```

s3ththompson:
@yoshuawuyts PTAL

PR #260, opened 2017-10-03
copy-only assets

yoshuawuyts:
Rewrote all of the asset handling logic; files are now served just in time. This means less memory usage, less chance of crashing, faster boot times and generally less chance of things going wrong.

To add a cherry on top; we now use `fs.copyFile` in the `bankai build` command to bypass userland memory copies entirely, which means building should be about as fast as we can get.

Hope this good; lmk if this works for y'all! :sparkles:

cc/ @s3ththompson @jondashkyle


index.js
```diff
@@ -182,14 +184,15 @@ Bankai.prototype.serviceWorker = function (cb) {
...
+  assert.equal(typeof filename, 'string')
   assert.equal(typeof cb, 'function')
   var stepName = 'assets'
   var self = this
   this.queue[stepName].add(function () {
-    var data = self.graph.data[stepName][edgeName]
-    if (!data) return cb(new Error('bankai.asset: could not find a file for ' + edgeName))
+    filename = path.join(self.dirname, filename)
+    var data = self.metadata.assets[filename]
+    if (!data) return cb(new Error('bankai.asset: could not find a file for ' + filename))
```
s3ththompson: Why is filesystem metadata being returned here?
Shouldn't this return just `filename`? See above where `http.js` expects to get a filename that it can serve.

Suggested fix:
```js
filename = path.join(self.dirname, filename)
cb(null, filename)
```

http.js
```diff
@@ -160,12 +161,12 @@ function start (entry, opts) {
...
     var prefix = 'assets' // TODO: also accept 'content'
     var name = prefix + '/' + params[1]
-    compiler.assets(name, function (err, node) {
+    compiler.assets(name, function (err, filename) {
       if (err) {
         res.statusCode = 404
         return res.end(err.message)
       }
-      res.end(node.buffer)
+      send(filename, res)
```
s3ththompson: I believe this should be:

```js
send(req, filename, {index: false}).pipe(res)
```

Flet:
OK! I made changes suggested by @s3ththompson 

I also moved the `tracker` instance into the `state` which fixes the test issues! 🎉

yoshuawuyts:
Alright, let's merge!

s3ththompson:
woot!

PR #261, opened 2017-10-03
fix document test

yoshuawuyts:
Document test were regressing; fixes that. 

- rel https://github.com/choojs/bankai/pull/259


yoshuawuyts:
This should be good; merging!

PR #264, opened 2017-10-03
read theme_color from manifest.json

### LINKED ISSUE #263: Theme color isn't derived correctly
The theme color is derived from the `color`-property in the `manifest.json`which isn't specified.
I suggest the `theme_color`-property should be used to set the content of the `theme-color` meta-tag.

f-mer:
Fixes #263


f-mer:
You are welcome!  😄  Always nice to make a contribution to such a cute project.

PR #266, opened 2017-10-04
Bundle splitting

goto-bus-stop:
now using [`split-require`](https://github.com/goto-bus-stop/browserify-dynamic-import/pull/4), a commonjs bundle splitting function and browserify plugin. see the linked PR and some of the comments below for discussion about `import('./whatever')` vs. `splitRequire('./whatever')` syntax

An example:
(this is also on glitch at https://glitch.com/edit/#!/faithful-turner?path=index.js:1:0 )

```js
var html = require('choo/html')
var app = require('choo')()
var splitRequire = require('split-require')

app.route('/', mainView)
app.mount('body')

var component
function mainView () {
  return html`<body>
    ${component ? component.render() : html`
      <h1>Loading</h1>
    `}
  </body>`
}

splitRequire('./SomeComponent', function (err, SomeComponent) {
  component = new SomeComponent()
  app.emitter.emit('render') // sshhh don't access this `.emitter` in a real app
  // this entire call should be in a store!
})
```

the below is outdated.

---

This is an initial exploration/take on adding the [`browserify-dynamic-import`](https://github.com/goto-bus-stop/browserify-dynamic-import) plugin to bankai, for easy bundle splitting. browserify-dynamic-import generates a main bundle, and dynamic bundles that are loaded at runtime. Use `import('./path/to/file')` to dynamically import a module. this could be used to lazy load code for routes or components.

There's still some work to be done on the plugin before it's 100% ready, like I have no clue what it'll do if you pass a dynamic path to `import()` for example, so. still need to work out some of the edge cases etc.

I had to disable SSR for now, because `import()` is unsupported syntax in node--i'm not sure what to do about this, i thought about adding a simple `require` hook that transforms `import()` server-side, but that might be too much magic

Todo:

- [x] Generate separate source maps using exorcist
- [x] Minify dynamic bundles
- [x] point bankai/http to dynamic bundles


goto-bus-stop:
Tried to address the server side thing :thinking:

I wrote a small package for this that replaces `import()` tokens with a
helper function that calls `require()`.
https://github.com/goto-bus-stop/dynamic-import-support

There's some other ways this could be done.

We could require that the app code supplies a polyfill for `import()`
for server side rendering if it needs it. It could use babel or
something to compile it first or just require `dynamic-import-support`
manually.

Or we could use `@std/esm` … https://github.com/standard-things/esm
That makes all of ES modules work. But then it'd be confusing that SSR
supports ES modules, while bundling doesn't—unless we add an ES modules
transform for browserify…

yoshuawuyts:
This is super neat work so far! – tested a few basic cases, and seems to work alright :D Once this lands we can start thinking about how to implement route-level support for it in Choo.

---

> We could require that the app code supplies a polyfill for import()
for server side rendering if it needs it. It could use babel or
something to compile it first or just require dynamic-import-support
manually.

What I wonder is if we actually need to support it. The main use case imagine for this would be to load modules and views asynchronously – as long as people don't need them to work in Node it should be good. I'm thinking mostly patterns such as [react-loadable](https://github.com/thejameskyle/react-loadable) and async routes at an application level would be used. 

I think we can safely ignore it for now – and perhaps special case the error in `lib/tty-error` to prettify the output, and explain why we can't support it for now.

goto-bus-stop:
status: this is not playing well with tinyify right now because of ordering issues. tinyify wants to add transforms _last_, for minification, and dynamic-import wants to add a transform first, for `import()` support. so dynamic-import has to run before tinyify. tinyify also assumes it is the first to add a plugin that alters the `pack` pipeline label, though, so tinyify has to run before dynamic-import. :') both plugins need to be more flexible about ordering ideally.

yoshuawuyts:
@goto-bus-stop Thanks for the update! Excited to have this working at some
point (perhaps Bankai 9.1; so no rush at all!) :D

goto-bus-stop:
`import()` syntax is proving to be quite difficult because most existing transforms don't support it. Acorn doesn't support it yet, so basically only transforms that use babylon work. the sequencing is difficult too for that reason. the dynamic import transform included with `browserify-dynamic-import` needs to run _before_ things like yo-yoify, but _after_ things like babelify.

i still kinda wish there was a commonjs way for doing this! on the server you can just `require()` things lazily when you need them, so it makes sense that commonjs hasn't spec'd something.

maybe some custom node-style thing that works on both the server and client would be a better way of going about this. there could still be a separate transform to convert `import()` calls to this node-style thing, with warning signs about being spec-noncompliant. I'm imagining something like:

```js
var splitRequire = require('split-require')

splitRequire('./a', function (err, exports) {
})
// instead of `import('./a').then(function (exports) {})`
```

on the server that could just be
```js
module.exports = function splitRequire (path, cb) {
  process.nextTick(function () {
    try { cb(null, require(path)) }
    catch (err) { cb(err) }
  })
}
```

and on the client it could be split off into a separate dynamically loaded bundle.

goto-bus-stop:
I think I'll change this over to the `split-require` style described in the previous comment, and have a separate transform that turns `import()` calls into `split-require` calls. That way people who want to be standards compliant or who want SSR can use `split-require` directly, but people who like ES modules, who maybe use transpiled ES modules, can use the `import()` syntax.

yoshuawuyts:
@goto-bus-stop this doesn't seem to be causing any errors on my machine at the moment. Can we merge this, and continue working on while it's deployed?

goto-bus-stop:
It currently relies on some unpublished versions of tinyify and split-require. Ill publish those tonight, then we can merge this!



On 14 November 2017 15:42:36 CET, Yoshua Wuyts <notifications@github.com> wrote:
>@goto-bus-stop this doesn't seem to be causing any errors on my machine
>at the moment. Can we merge this, and continue working on while it's
>deployed?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/choojs/bankai/pull/266#issuecomment-344279740

-- 
Sent from mobile. Please excuse my brevity.

yoshuawuyts:
@goto-bus-stop should we add a quick test to make sure bundle splitting works? E.g. just make sure the bundle is actually created, and is not empty?

goto-bus-stop:
added a test. looking into making `bankai inspect` work with this too, it looks like `disc` accepts multiple bundles as input so it would be neat to get that going.

yoshuawuyts:
Yay! — is there anything left to do? I think we could leave the disc bits for now if needed C:

goto-bus-stop:
Yeah I disabled splitting in `inspect` for now, because supporting fullPaths in split-require is going to take some time. I uhh. ..think this is good to go then! 😲



On 17 November 2017 16:43:15 CET, Yoshua Wuyts <notifications@github.com> wrote:
>Yay! — is there anything left to do? I think we could leave the disc
>bits for now if needed C:
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/choojs/bankai/pull/266#issuecomment-345278752

-- 
Sent from mobile. Please excuse my brevity.

PR #268, opened 2017-10-05
pass in opts to http interface

lrlna:
currently the http interface doesn't pass in the opts. this PR adds the `reload` opt to `opts` that a user might also pass in. 

Thanks ^___^

### DIFF
--- a/http.js
+++ b/http.js
@@ -27,7 +27,8 @@ function start (entry, opts) {
   assert.equal(typeof opts, 'object', 'bankai/http: opts should be type object')
 
   var quiet = !!opts.quiet
-  var compiler = bankai(entry, { reload: true })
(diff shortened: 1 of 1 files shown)


http.js
```diff
@@ -27,7 +27,8 @@ function start (entry, opts) {
   assert.equal(typeof opts, 'object', 'bankai/http: opts should be type object')
 
   var quiet = !!opts.quiet
-  var compiler = bankai(entry, { reload: true })
+  opts.reload = true
+  var compiler = bankai(entry, opts)
```
yoshuawuyts: Could we do:

```js
Object.assign({ reload: true }, opts })
```

lrlna: oh yea, sure! is this ok, or do you want it inline?

PR #270, opened 2017-10-06
fix viewport location

yoshuawuyts:
Firefox was getting really upset at this. Solves the linked issue:

- https://stackoverflow.com/questions/24875520/the-character-encoding-declaration-of-the-html-document-was-not-found


goto-bus-stop:
if people configure documentify transforms in package.json, they could add a stylesheet or something in the head before bankai's transforms run, which would push the meta tag out of the first 1024 bytes again. maybe `viewportTransform` could use hyperstream's `_prependHtml` instead of `_appendHtml`?

goto-bus-stop:
now #300 :D

yoshuawuyts:
Lmao — ooops :D

PR #274, opened 2017-10-09
add module deps

yoshuawuyts:
Adds an overview of all used modules in the project, and their respective locations. Done on request of @jasonlaster so browsers like FireFox can figure out more about the code (e.g. walk it themselves).

The idea is to add an extra comment at the end of the bundle similar to source maps to indicate where to retrieve the module deps. This isn't part of this PR yet, because we're not completely sure what that would look like.

Furthermore this is also something that can later be adopted by other bundlers such as WebPack.

---

cc/ @jsonlaster could you link to the related issues in FireFox so we can add some more context? Thanks! :grin:

### DIFF
--- a/http.js
+++ b/http.js
@@ -170,6 +170,17 @@ function start (entry, opts) {
     })
   })
 
+  router.route(/\/module-deps\.json/, function (req, res) {
+    compiler.moduleDeps(function (err, json) {
+      if (err) {
+        res.statusCode = 404
+        return res.end(err.message)
+      }
+      res.setHeader('content-type', 'application/json')
+      gzip(json, req, res)
+    })
+  })
+
   router.route(/\/reload/, function sse (req, res) {
     var connected = true
     emitter.on('scripts:bundle', reloadScript)
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var assert = require('assert')
 var path = require('path')
 var pino = require('pino')
 
+var moduleDeps = require('./lib/module-deps')
 var localization = require('./localization')
 var queue = require('./lib/queue')
 
@@ -115,6 +116,7 @@ function Bankai (entry, opts) {
     }
   })
 
+  this.entry = entry
   this.metadata = this.graph.metadata
 }
 Bankai.prototype = Object.create(Emitter.prototype)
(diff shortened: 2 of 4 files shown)


jasonLaster:
This is fantastic!

We are talking about linking modules in the debugger #4333

goto-bus-stop:
is this only for getting access to full file paths? if so, couldn't they live in the sourcemap file? source maps include a `sources` array with relative file names—i could imagine an `x_full_paths` array with full paths to each file being added during development.

yoshuawuyts:
Fair point — that would also remove the need to include an extra comment
that points to the routes file.

Is there a source map spec anywhere? We should probs make sure that it
doesn't prohibit extending the mapping with custom values

jasonLaster:
the spec is here:

https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?pli=1#heading=h.535es3xeprgt

It would be interesting to see something added to the source maps spec.  

* I'm not sure if `sourceRoot + source` give you the full_path for a source.
* it would be interesting to capture the source dependencies perhaps `x_source_deps`

goto-bus-stop:
I think sourceRoot normaly points to a web accessible version of the source files, for when you're not including source files in the `sourcesContent` property. Maybe setting it to a file:// url during development could work tho. But if we're going to use `bankai://` in the sources array, webpack-style, concatenating sourceRoot+source would not give a full file path because of the protocol.

Something to consider for the sourceDepsURL comment is that most tools already take the sourcemapurl comment into account. A new comment could be removed by uglify, or moved around by Babel, etc, unless it is added after all other transformations have run, or every tool is taught about it. Extending the source map can also have that problem unfortunately if some tool discards input sourcemaps and just adds its own.

yoshuawuyts:
From IRC:
```txt
16:14 <yoshuawuyts> goto-bus-stop: do you have an idea on how to resolve https://github.com/choojs/bankai/pull/274 ?
16:18 <goto-bus-stop> yoshuawuyts: hmm it's kinda difficult because it seems like, whether it's a new comment or a new sourcemap key, all or most existing tools will discard it during transforms
16:19 <yoshuawuyts> goto-bus-stop: :(
16:19 <yoshuawuyts> goto-bus-stop: so extending the spec, is, like, hard
16:19 <goto-bus-stop> yoshuawuyts: yaeh
16:19 <goto-bus-stop> yoshuawuyts: one option i can think of is to add it at the very end after all transforms have run
16:20 <yoshuawuyts> goto-bus-stop: that wouldn't be too bad!
```

yoshuawuyts:
Don't think we can do this at the moment. It seems we agree doing it in the sourcemaps is the right approach; but we'd need to update quite a bit of tooling for it to happen. Closing for now!

PR #276, opened 2017-10-10
Fix minify source mapping, fixes #275

### LINKED ISSUE #275: Source maps are broken
We had source maps working, but they broke again. @goto-bus-stop any ideas of what might have happened?

goto-bus-stop:
I'm not sure what was causing it, I think maybe the `sourcesContent` overrides
that were in there previously? Then because the source files are not being
served either, the browser doesn't know what sourcemapped files look like.

Whatever it was, using uglify's built-in inline source map detection seems to
work :tada:

### DIFF
--- a/lib/minify.js
+++ b/lib/minify.js
@@ -1,7 +1,6 @@
 var convert = require('convert-source-map')
 var uglify = require('uglify-es')
 var assert = require('assert')
-var path = require('path')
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Yay, this works! :grin:

yoshuawuyts:
`v9.0.0-rc4`

PR #277, opened 2017-10-10
Enable ascii_only output to prevent utf8 mangling

toddself:
Uglify seems to have a bad time at parsing/handling UTF-8 characters, and the
`ascii_only: false` default is the culprit.

Switching this on causes the UTF-8 garbage to disappear!


toddself:
[From the docs](https://github.com/mishoo/UglifyJS2/tree/harmony#output-options):

> ascii_only (default false) -- escape Unicode characters in strings and regexps (affects directives with non-ascii characters becoming invalid)

PR #283, opened 2017-10-13
full reset and render on terminal resize

Flet:
Before:
before

After:
after

The `console.reset()` feels a little hacky, but `differ` does not seem to deal with the extra cruft when resetting/clearing. Maybe this is good enough for now?


Flet:
Closing in favor of #284

PR #284, opened 2017-10-13
switch to ansi-diff and make resizing nice

Flet:
I like this a lot better than #283 :)

Very Smooth
smooth


Flet:
Vertical resizing does not seem to be covered :(

Flet:
updated commit... now vertical resizing should work nicely.

smoothvertical

PR #288, opened 2017-10-16
add document description meta from manifest

bendik:
any feedback welcome 🛄


Flet:
Can you add `'manifest:description'` to the dependencies for the document node in `index.js`?
https://github.com/choojs/bankai/blob/master/index.js#L94

This will ensure the description is available before the document node runs.

LGTM once this is done :)

yoshuawuyts:
Ah yeah; this is neat! — could you perhaps rebase so we can merge? Should be fairly straight forward (but I can help if you need any). Thanks heaps!

PR #299, opened 2017-10-17
pass ssr error in callback - ensures SSR will re-render after errors

Flet:
I noticed that SSR would never fire again after having an error. This is because the `ssrError` attribute hanging off of `render` was never cleared between runs, so the logic assumed every subsequent run was an error.

This PR updates the render function to pass the SSR error back in the callback instead of hanging it off of `render`. This ensures SSR works after errors are fixed in the app code and also ensures safety if multiple renders are firing at once for some reason as the attribute is not shared.

### DIFF
--- a/lib/server-render.js
+++ b/lib/server-render.js
@@ -19,6 +19,7 @@ function serverRender (entry, cb) {
   assert.equal(typeof entry, 'string')
 
   var app
+  var ssrError = null
   var channel = new EventEmitter()
   channel.once('list', function (list) {
     render.list = list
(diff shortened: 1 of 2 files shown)


lib/server-render.js
```diff
@@ -66,7 +67,7 @@ function serverRender (entry, cb) {
     }
   }
 
-  cb(null, render)
+  cb(null, render, ssrError)
```
goto-bus-stop: is there a reason the error can't be passed in the first parameter?

yoshuawuyts: Agree — think it'd be nice if `error` had an extra property on it so we can figure out what type of error it is (pretty multiline one, regular unknown error, SSR error) and we proceed from there. Prevents from adding more parameters in the future too, as we expand our error handling capabilities (:

Flet: makes total sense! I was overthinking this :)

Flet:
Updated to use callback error instead of additional arg.

PR #300, opened 2017-10-17
fix document viewport tag

yoshuawuyts:
Makes it so the encoding is in the first 1024 bytes of the document, which should improve parsing speed and prevent FF from warning. Thanks!


yoshuawuyts:
This is fairly straight forward, don't think anyone will object — merging!

PR #301, opened 2017-10-18
Subresource integrity

yoshuawuyts:
Add subresource integrity checks — should (partially) help mitigate a certain class of MITM attacks. Also bumped to `buffer-graph@4` which now uses sha512, which in turn we can use to set the integrity checks.

## See Also
- https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

## Todo
- [x] make sure the stores URLs in the `service-worker` globals are stores in hex
- [x] store the hashes in `bankai build` are written as hex
- [x]  fix tests that rely on hashes


yoshuawuyts:
Had to disable this for CSS; conflicted with preloading. See https://twitter.com/yoshuawuyts/status/920794607314759681 :(

lib/graph-service-worker.js
```diff
@@ -114,9 +114,9 @@ function find (rootname, arr, done) {
 function fileEnv (state) {
   var script = [
     'https://cdn.polyfill.io/v2/polyfill.min.js',
-    `${state.scripts.bundle.hash}/bundle.js`
+    `${state.scripts.bundle.hash.hexSlice().slice(0, 16)}/bundle.js`
```
goto-bus-stop: afaik `hexSlice` is not public API, maybe better to use `.toString('hex', 0, 16)`?

goto-bus-stop:
LGTM otherwise! very nice :D

lib/graph-service-worker.js
yoshuawuyts: Updated to use `.toString()`! Kept the `.slice()` method in there because we had to do `.toString('hex', 0, 8)` which isn't as clear that it returns 16 bytes.

yoshuawuyts:
Implemented feedback. Merging! :D Yayayay

PR #302, opened 2017-10-18
Update documentify

goto-bus-stop:
this changes the order transforms are run in: those defined in
`package.json` are now run first, like in browserify. technically
that's a breaking change, but maybe we can merge this without bumping
bankai because it's still `-rc`? :P

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "css-extract": "^1.2.0",
     "debug": "^3.0.1",
     "disc": "^1.3.2",
-    "documentify": "^2.0.0",
+    "documentify": "^3.0.0",
     "exorcist": "^1.0.0",
     "explain-error": "^1.0.4",
     "factor-bundle": "^2.5.0",


Flet:
@goto-bus-stop Thanks! this might help solve issues with folks who want to `babelify` or `tsify` :)

Flet:
ah nope :(  `babelify` and `yo-yoify` still don't seem to get along.

goto-bus-stop:
yeah this only changes how the html is transformed. i guess the problem with babelify+yoyoify is that babel transforms template strings before yoyoify finds them? :/ package.json transforms run before transforms included with bankai. one thing you _could_ do is configure babel to use [babel-plugin-yo-yoify](https://github.com/goto-bus-stop/babel-plugin-yo-yoify) haha.

Flet:
ah nice! I'll use that!

yoshuawuyts:
We might eventually pull in Babel for a two-bundle system using module
script tags + nomodule. That way we can compile in whatever older browsers
don't have, without impacting newer browsers. Just a thought at the moment,
but might be neat in the future (:

PR #303, opened 2017-10-19
Manually blow away cache when server rendering

goto-bus-stop:
Proxyquire normally only removes the required file from the cache, but
not its dependencies. It removes everything if you use proxyquire to
stub a global variable, but I think stubbing a variable would be a weird
thing for bankai to do :thinking:

This patch deletes the entire module cache before requiring the app
entry point. It keeps native modules cached because requiring those
twice doesn't work. At the end the cache is restored.

This should solve the issue where changes in non-entry modules were not
picked up by server rendering.

Most of the existing modules try to be very smart about what they delete
from the cache, but I think this should also work?

### DIFF
--- a/lib/server-render.js
+++ b/lib/server-render.js
@@ -1,5 +1,4 @@
 var EventEmitter = require('events').EventEmitter
-var proxyquire = require('proxyquire')
 var assert = require('assert')
 
 var getAllRoutes = require('wayfarer/get-all-routes')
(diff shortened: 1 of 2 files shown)


goto-bus-stop:
@Flet could you try if this works for you?

Flet:
OK, trying this now. Do you think it makes sense to blow away only items with keys in the target application's basedir? This might help avoid reloading bankai things?

goto-bus-stop:
hmm maybe, but if bankai is installed locally it'll also be in the application's basedir. i don't think applications would usually require any bankai things, and this reinstates the cache after SSR-ing, so bankai modules shouldn't be reloaded.

PR #305, opened 2017-10-21
reload on document change instead of script

Flet:
With the addition of the subresource integrity logic, we're seeing issues when running `bankai start`. When the bundle changes the page gets reloaded, but sometimes the document has not been yet been rebuilt with the new hash on the `<script>` tag. This causes a `subresource integrity error` to be thrown in browser and the script fails to load.

This PR changes `http.js` to only trigger a reload when the document has been fully rebuilt. This should avoid the race condition.

It feels a little bad not renaming everything to `reloadDocument` all the way down as it would be a bit more accurate... I'm happy to update this if this PR is going in the right direction!

### DIFF
--- a/http.js
+++ b/http.js
@@ -80,7 +80,7 @@ function start (entry, opts) {
     }
     state.files[nodeName] = data
 
-    if (name === 'scripts:bundle') emitter.emit('scripts:bundle', node)
+    if (name === 'documents:index.html') emitter.emit('documents:index.html', node)
     if (name === 'styles:bundle') emitter.emit('styles:bundle', node)
 
     // Only calculate the gzip size if there's a buffer. Apparently zipping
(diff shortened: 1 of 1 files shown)


http.js
```diff
@@ -203,7 +203,7 @@ function start (entry, opts) {
     function disconnect () {
       clearInterval(interval)
       if (connected) {
-        emitter.removeListener('scripts:bundle', reloadScript)
+        emitter.removeListener('documents:index.html', reloadScript)
```
yoshuawuyts: This should probably be `documents:list`

yoshuawuyts: oh actually — nevermind; this doesn't seem to work the way I'd hoped it would :/

yoshuawuyts:
Seems to be conflicting with #317 :( — would you care to update? We should probably merge this! :grin:

yoshuawuyts:
Alright, rebased. Waiting for tests to pass, and we should be good to merge! :tada:

yoshuawuyts:
Tests passed. Merging!

PR #314, opened 2017-10-30
delete the ‘:/’ property before return an array

ZhouHansen:


yoshuawuyts:
This makes a lot of sense. Thanks!

PR #316, opened 2017-10-31
pluralize style -> styles

yoshuawuyts:
Final breaking change for the API. Changes `.style()` to `.style()`. Allows us to experiment in the future with splitting CSS bundles, and the like. Not sure what the approach would be, but allows for this to happen without breaking the API. Thanks!


yoshuawuyts:
cool! — merging :D

PR #317, opened 2017-11-01
watch fixes

Flet:
I think I found a source for "warning: possible EventEmitter memory leak detected" messages. It looks like a watcher in `graph-service-worker.js` was being added every reload.

Also, I noticed that `watchDirs` regexes were we being generated as `^assets` instead of the full absolute path `^/Users/flet/code/stuff/assets`.

I also added some additional debug un `utils` and updated it to use `debug`s formatters to avoid unnecessary string concats if debug is off.


yoshuawuyts:
Neat! :D

PR #318, opened 2017-11-02
remove duplicated sha512 and fix typo

ZhouHansen:


goto-bus-stop:
nice catch! it's strange that the reload code was still being executed even though the integrity value was wrong--maybe that's only checked for remote resources?

ZhouHansen:
I red [https://www.w3.org/TR/SRI/](https://www.w3.org/TR/SRI/) , It doesn't say that's only checked for remote resources, but only for integrity checks if they are same-origin or are the result of explicit access granted to the loading origin via Cross Origin Resource Sharing.

And see [use-casesexamples](https://www.w3.org/TR/SRI/#use-casesexamples), maybe it's unnecessary to use integrity at development mode.  Also see [subresource-integrity-and-performance](https://stackoverflow.com/questions/36093920/subresource-integrity-and-performance).

PR #320, opened 2017-11-03
sorta fix http2

### LINKED ISSUE #313: node v8.8 - HTTP2 issue?
Having some issues with upgrade to node 8.8

Bankai falls over when a website reloads ... I don't get much more in the logs other than an ELIFECYCLE error.

```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! descartes@1.0.0 dev: `bankai start index.js`
npm ERR! Exit status 1
```

yoshuawuyts:
As per https://github.com/nodejs/node/issues/16666, this sorta fixes http/2 support. It's a step up from where we're at right now. If this turns out to be too unstable, we should disable http2 by default for now, and perhaps make it opt-in until it further stabilizes, but I suggest we wait and see for a bit.

Closes #313. Thanks!

### DIFF
--- a/http.js
+++ b/http.js
@@ -189,17 +189,13 @@ function start (entry, opts) {
     res.write('retry: 10000\n')
 
     var interval = setInterval(function () {
+      if (res.finished) return // prevent writes after stream has closed
       res.write(`id:${id++}\ndata:{ "type:": "heartbeat" }\n\n`)
     }, 4000)
 
-    // Attach an error handler, but no need to actually handle the error.
-    // This is a bug in Node core according to mcollina which will be fixed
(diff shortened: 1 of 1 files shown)


http.js
```diff
@@ -192,14 +192,9 @@ function start (entry, opts) {
       res.write(`id:${id++}\ndata:{ "type:": "heartbeat" }\n\n`)
```
apapirovski: Just quoting myself :)

> you could add a little extra safety check to write that only fires if !res.finished.

yoshuawuyts: That's a great idea; updating!

PR #323, opened 2017-11-06
remove polyfill.io

yoshuawuyts:
Removes the polyfill.io transform service. I don't feel comfortable pushing this to the `9.0.0` stable release, and perhaps we should consider adding this on a later stage. Right now it makes local development harder (especially offline), does not work well with content integrity checks, and will partially be mitigated if we enable a transpiler. Let's chat on IRC if we need this back at some point, but think for the initial release we should leave it out.

Hope this makes sense; thanks!


Flet:
I liked the idea of it! Probably fine to remove since it wasn't art of `bankai@8` :)

Folks who want to have some guarantee of compatibility may be disappointed, especially since using babelify with bankai is not quite painless right now.

Overall 👍

PR #326, opened 2017-11-08
set content-type via MIME type of asset name

Flet:
I was running into issues when trying to serve .svg files from /assets in bankai. It turns out that the content-type was not set correctly.

This PR adds the [`mime` package](https://github.com/broofa/node-mime) (~2.8KB) and uses it to determine what content-type to set based on the file name.


Flet:
Closing as this should be covered by #260

PR #328, opened 2017-11-13
Update yo-yoify

goto-bus-stop:
changes:
- removes support for onload.
- adds support for `<div ${object} />` splat properties.
- generates smaller bundles for choo by removing `choo/html` requires.
- compiles babelified template strings too


Flet:
You're the best @goto-bus-stop :)

PR #330, opened 2017-11-14
allow curl from CLI

yoshuawuyts:
Allows using `curl(1)` from the command line to quickly test output. Only available on the command line, so no security issues here (I mean; we're self-signing certs here, so you don't wanna run it in prod anyway).

Also added a default body for the `301` redirect, so folks will have a better time if they ever hit it. Thanks!


yoshuawuyts:
> 17:08 <goto-bus-stop> the curl thing is a clever idea!

Welllp, good enough for me :P

Issue #329, opened 2017-11-14
Babelify config for Browserify does not work with rc10


kahlil:
I use the following Browserify config:

```js
"browserify": {
	"transform": [
		[
			"babelify",
			{
				"presets": [
					"env"
				]
			}
		]
	]
}
```
And Bankai can't parse the `import` syntax:

```
Failed while processing 'scripts'.

 --> index.js:2:1
  |
2 | import { html, render } from 'lit-html';
  | ^ Syntax Error

Hmmm. We're having trouble parsing a file.
```
The necessary modules are installed.

yoshuawuyts:
Oh, that's odd. I would've expected https://github.com/shama/yo-yoify/pull/58 by @goto-bus-stop to have solved our final compat issues. My guess is there's probably some parser in the tree that's operating on an outdated version.

Also, on a side note we're going to be baking in Babel support for stable language features sometime in the next 2 weeks. Hopefully that'll fix this issue in the long run C:

kahlil:
👍  for baking Babel in! Still this simple Browserify config should still work, right?

yoshuawuyts:
Yeah, I would've expected it to! Not sure why it isn't parsing the `import` syntax. [Here's a project where we're using browserify transforms too](https://github.com/yoshuawuyts/millennial-js/blob/master/package.json#L11-L15). So I don't think it's being glossed over, ghmmm. I'm still suspecting a transform using an outdated parser somewhere.

Could you perhaps run `bankai build` to see if it crashes with a different error? Thanks!

kahlil:
This is the output on `bankai build`: 

```
❯ npx bankai build
14:30:28 ✨  Compiling & compressing files

14:30:28 ✨  created: dist
14:30:28 🚨  bankai.asset: could not find a file for 
14:30:28 ✨  created: dist/manifest.json
14:30:28 ✨  created: dist/manifest.json.gz
14:30:28 ✨  created: dist/manifest.json.deflate
14:30:28 ✨  created: dist/manifest.json.br
14:30:28 🚨  Failed while processing 'scripts'.

 --> index.js:2:1
  |
2 | import { html, render } from 'lit-html';
  | ^ Syntax Error

Hmmm. We're having trouble parsing a file.
```

yoshuawuyts:
Ugh; looks like we did _too good_ of a job on prettifying errors >.>

Let me try and repro this. Be back with you in a bit!

kahlil:
thanks!

[...]

PR #331, opened 2017-11-14
fix brfs

yoshuawuyts:
Fixes the parsing issue Bankai was having https://github.com/choojs/bankai/issues/329! — e.g. allows `babel` to run from the `package.json`. Unfortunately it can't quite run as a global transform, but that's a separate issue. Thanks!

P.s. We should still perform the patches as mentioned in https://github.com/choojs/bankai/issues/329#issuecomment-344269383.


yoshuawuyts:
Swoosh!

PR #334, opened 2017-11-19
rm stdout-stream

ZhouHansen:
I think stdout-stream is needless when use node which version is newer than v0.12.x. The LTS Status of v0.12.x and v0.10.x is End-of-Life, neither support es6. You can't run bankai with for example v0.12.17, because it doesn't support tagged template literals, that means you should install node which version is newer than v0.12.x to run bankai. Through my test newer version of node had solved the problem which stdout-stream solved, there is no blocking when `process.stdout` being piped.

* [nodejs/Release](https://github.com/nodejs/Release) 
* [node.green](http://node.green/)


yoshuawuyts:
thanks!

Issue #341, opened 2017-11-23
Babel requires preset


amsross:
`bankai start` returns this error:
```
scripts:browserify.bundle Couldn't find preset "es2015" relative to directory "/app/client/node_modules/hawk" while parsing file: /app/client/node_modules/hawk/dist/browser.js
Error: Couldn't find preset "es2015" relative to directory "/app/client/node_modules/hawk" while parsing file: /app/client/node_modules/hawk/dist/browser.js
    at /app/client/node_modules/babel-core/lib/transformation/file/options/option-manager.js:293:19
    at Array.map (<anonymous>)
    at OptionManager.resolvePresets (/app/client/node_modules/babel-core/lib/transformation/file/options/option-manager.js:275:20)
    at OptionManager.mergePresets (/app/client/node_modules/babel-core/lib/transformation/file/options/option-manager.js:264:10)
    at OptionManager.mergeOptions (/app/client/node_modules/babel-core/lib/transformation/file/options/option-manager.js:249:14)
    at OptionManager.init (/app/client/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/app/client/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/app/client/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/app/client/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at Babelify._flush (/app/client/node_modules/babelify/index.js:26:24)
```

By adding `babel-presets-es2015` (`npm install --save-dev babel-presets-es2015`), I was able to successfully start the dev server.

goto-bus-stop:
Hmm, it should not try to transform code from node_modules with custom presets, only with babel-preset-env. I think we need to add `babelrc: false` to the babelify config in `lib/graph-script.js`, so that .babelrc files in node_modules are not picked up.

amsross:
@goto-bus-stop that addresses the issue. I'll take a look at why that works and open a PR for it.

Thanks!

toddself:
There is something more going on here and we might have to think about this some more.

Example: [diff](https://github.com/kpdecker/jsdiff), when you install from npm has a re-built version to use in the browser.

This is build, in part, by using babel.  So, to configure babel, there is a `babel` section in the package.json, but the presets are stored as devDepenendencies (correctly since they're not needed for the installed version).

[...]

PR #342, opened 2017-11-23
Ignore .babelrc files in babelify

amsross:
The goal is to ignore `.babelrc` files in `node_modules` per @goto-bus-stop's [comment](https://github.com/choojs/bankai/issues/341#issuecomment-346684897) in [#341](https://github.com/choojs/bankai/issues/341) to solve the issue raised there.

### DIFF
--- a/lib/graph-script.js
+++ b/lib/graph-script.js
@@ -54,6 +54,7 @@ function node (state, createEdge) {
   b.transform(glslify)
   b.transform(babelify, {
     global: true,
+    babelrc: false,
     presets: [
       [babelPresetEnv, {
         targets: { browsers: browsers }


amsross:
I don't have a lot of experience with babel, or, specifically `babelify`, but it seems to me that this would remove the possibility of using `.babelrc` to determine the behavior of babel in/on your own project. Is this not the case here?

yoshuawuyts:
It seems there's  few more comments in #341; not quite sure yet what the right solution is, as it seems this patch would have tradeoffs, right?

amsross:
Correct. I think this would have negative consequences based on expectations set in the docs.

amsross:
Out of my depth; Closing in favor of whatever comes of #341

Issue #292, opened 2017-10-17
EADDRINUSE silent fail


soyuka:
Hi,

I've nginx setup on the port `8080`. When I use `bankai start` it silently fails (`onexit` doesn't check if the argument is an error, which is given by `uncaughtException` for example) with a code of `0`.

I'm not sure why `getPort` (used in `lib/cmd-start`) doesn't give the correct port, access rights are correct. Looks like we're going through [this line](https://github.com/mikeal/getport/blob/master/index.js#L21) in my context. You may want to switch to [`get-port`](https://github.com/sindresorhus/get-port) instead?

PR #344, opened 2017-11-24
Use get-port instead of getport

soyuka:
This fixes my issue.

If you want me to add a test lmk.

### DIFF
--- a/lib/cmd-start.js
+++ b/lib/cmd-start.js
@@ -1,4 +1,4 @@
-var getPort = require('getport')
+var getPort = require('get-port')
 
 var bankai = require('../http')
 var http = require('./http-server')
@@ -16,11 +16,13 @@ function start (entry, opts) {
     })
   })
 
-  getPort(8080, 9000, function (err, port) {
-    if (err) state.error = err
+  getPort({port: 8080}).then(function (port) {
     server.listen(port, function () {
       state.port = port
     })
+  })
+  .catch(function (err) {
+    state.error = err
   })
 }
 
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
Tests would be good! I'll take a look on Monday

soyuka:
Will add tests monday!

goto-bus-stop:
@soyuka sorry, what issue is solved by this? / what's the difference between those modules?

soyuka:
@goto-bus-stop Meh, dunno. I'm running nginx on the bankai default port and it crashes on launch :|. I made an issue about this last month. Lmk if you want me to rebase!

goto-bus-stop:
Ohh I see, it's #292. I think this should be fine, will check in the afternoon

goto-bus-stop:
oh, so get-port tries to create a server while getport tries to connect to a port. i don't know when that would yield different results but if it does it makes more sense for bankai to use the former. thanks for the rebase, sorry this had to wait so long!

it looks like get-port returns a Promise and doesn't accept a callback though, could you update that? and also get rid of the package-lock please :) thanks 🙏

soyuka:
Done, don't worry about the waiting time haha

lib/cmd-start.js
```diff
@@ -16,7 +16,7 @@ function start (entry, opts) {
     })
   })
 
-  getPort(8080, 9000, function (err, port) {
+  getPort({port: 8080}).then(function (err, port) {
```
goto-bus-stop: just one last thing: these should be `.then(function (port) { }).catch(function (err) {})`

soyuka: I should not try to fix things while in holidays \o/

goto-bus-stop:
yay thanks!! 🎉

PR #349, opened 2017-12-03
Enable tinyify in `bankai inspect`

goto-bus-stop:
tinyify 2.4.0 detects the browserify `fullPaths` option and produces a
bundle that `disc` can use.


yoshuawuyts:
@goto-bus-stop rad!!!!

Issue #346, opened 2017-11-27
Stylesheet not loading in latest Safari


jondashkyle:
It looks like setting the `rel="stylesheet"` attribute onload does not work in the most recent Safari. Changing this to simply declare `rel` by default correctly loads the stylesheet, but am unsure if preloading works as intended.

```
<link rel="preload" as="style" href="/6b2bede60756eb41/bundle.css" rel="stylesheet">
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#The_basics

yoshuawuyts:
Yeah, i think we are trying to redeclared this.rel. perhaps we need to clear it first?

jongacnik:
perhaps useful, this is an example of the type of async css we're trying to achieve. It runs in most recent Safari, but that is because there is a polyfill in the source

https://master-origin-loadcss.fgview.com/test/preload.html

greggraf:
@yoshuawuyts - it looks like the polypill is already in there. It just appears that the script block is not getting processed, and I think its the `nomodule` and `integrity` attributes on the generated `script` tag.

when I remove them from https://github.com/choojs/bankai/blob/master/lib/graph-document.js#L111 locally, the polypill works for me and my css loads in mobile safari.

I don't think those attributes belong on an inline script tag. I will put together a pull request.

gavinpatkinson:
this still appears to be an issue

yoshuawuyts:
@greggraf this issue should now be fixed, right? :D

greggraf:
@yoshuawuyts - when I check with Bankai v9.2.1 on desktop Safari I see the css getting requested in the network tab and rendered on my page.

goto-bus-stop:
:100: thanks!

PR #350, opened 2017-12-06
Remove attributes from loadCSS polyfill inline script

greggraf:
Update tests and code to remove attributes from loadCSS polyfill inline script

Inline script to polyfill link rel=preload behavior in browsers like Safari was not getting processed. After removing the `nomodule` and `integrity` attributes it is processed and `bundle.css` is loaded.

This work is to solve #346


goto-bus-stop:
thanks! does it work when you only remove the `integrity` attribute but keep the `nomodule` one? or do they both need to go? afaik `integrity` does nothing for inline scripts, but `nomodule` should still prevent that code from executing in new browsers.

test/document.js
```diff
@@ -86,7 +86,7 @@ tape('server render choo apps', function (assert) {
...
...ner("load",f),i.media=d||"all"}var g,h=a.document,i=h.createElement("link");if(c)g=c;else{var j=(h.body||h.getElementsByTagName("head")[0]).childNodes;g=j[j.length-1]}var k=h.styleSheets;i.rel="stylesheet",i.href=b,i.media="only x",e(function(){g.parentNode.insertBefore(i,c?g:g.nextSibling)});var l=function(a){for(var b=i.href,c=k.length;c--;)if(k[c].href===b)return a();setTimeout(function(){l(a)})};return i.addEventListener&&i.addEventListener("load",f),i.onloadcssdefined=l,l(f),i};"undefined"!=typeof exports?exports.loadCSS=b:a.loadCSS=b})("undefined"!=typeof global?global:this);;(function(a){if(a.loadCSS){var b=loadCSS.relpreload={};if(b.support=function(){try{return a.document.createElement("link").relList.supports("preload")}catch(b){return!1}},b.poly=function(){for(var b=a.document.getElementsByTagName("link"),c=0;c<b.length;c++){var d=b[c];"preload"===d.rel&&"style"===d.getAttribute("as")&&(a.loadCSS(d.href,d,d.getAttribute("media")),d.rel=null)}},!b.support()){b.poly();var c=a.setInterval(b.poly,300);a.addEventListener&&a.addEventListener("load",function(){b.poly(),a.clearInterval(c)}),a.attachEvent&&a.attachEvent("onload",function(){a.clearInterval(c)})}}})(this);</script>
```
goto-bus-stop: we can also get rid of the `__PRELOAD_INTEGRITY__` variable and replaces in this file

greggraf: In my testing, the `nomodule` attribute prevents the inline script from running in both Safari and Chrome. This behavior appears correct for when ES6 Modules (and the `nomodule` attribute) are supported.

For versions of Safari that support ES6 modules but not the `rel="preload"` resource hint, the script should run regardless of module support to polyfill the lack of preload support, so the `nomodule` attribute should not be added back.

greggraf:
In my testing, the `nomodule` attribute prevents the inline script from running in both Safari and Chrome. This behavior appears correct for when ES6 Modules (and the `nomodule` attribute) are supported.

For versions of Safari that support ES6 modules but not the `rel="preload"` resource hint, the script should run regardless of module support to polyfill the lack of preload support, so the `nomodule` attribute should not be added back.

goto-bus-stop:
> For versions of Safari that support ES6 modules but not the `rel="preload"` resource hint

gotcha, makes sense--this is the context i was missing :) thanks for explaining!

this LGTM then, if you could remove the remaining references to `__PRELOAD_INTEGRITY__` in the test file?

test/document.js
greggraf: Right! Done!

goto-bus-stop:
thanks! 🎉

yoshuawuyts:
Oh, by the way - I found out the other day how to do CSP for inline scripts :D https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

Issue #348, opened 2017-11-28
Add a --no-babelify switch


toddself:
As noted in #341, if we are running babelify globally, any package which uses babelify in it's build process but does not include the babel packages in it's `dependencies` list (since the package's published version is a pre-built bundle), will cause Bankai to fail.

The only way to discover this issue is to try to build something, if the offending package is the dependency of a dependency it gets even harder to figure out.  If it is expecting multiple presets or babel packages, each one has to be discovered independently (the stacktrace is only moderately helpful).

Also, as with the case for [diff](https://github.com/kpdecker/jsdiff), the version of babel/babelify is incompatible with the version used for building the library, so even installing the required babel packages in the top-level project causes the build to fail since the packages do not run under the babel runtime provided.

This means it is currently impossible to build a project which relies on jsdiff using bankai (or any other package which exhibits similar attributes).

Ergo, we need to provide some ability to tell bankai to _not_ run babelify globally (or at all) or we need to turn off the reading of custom `.babelrc` configurations.

We've discussed the latter of these options and that's I think an even more worse scenario than the first one.

I know the idea is to have a zero-configuration tool, but the problem here is that babel _requires_ configuration; it is impossible to use it without configuration, and it's configuration is definitely not a one-size-fits-all type of thing (or the tool itself would be infinitely less configurable).

yoshuawuyts:
Hey, thanks for the issue. Can I please stress that one of the primary goals of Bankai is to not introduce any configuration. I don't agree with the line of reasoning in the last paragraph, and reject the premise that because something lower-level has configuration, we should also have configuration. Programming is a game of tradeoffs, and it's up to us to find the right ones for Bankai.

I think we should definitely talk about what the best approach is on how to combine Bankai & Babel, but pushing for command line flags is counter-productive. If we are going to have this discussion, we should not go down the route of flags as the first solution to every problem.

To be frank, I'm getting a little tired (an possibly frustrated) with the interactions we've been having recently. I know we approach problems from different angles (which is great!), but it feels that our interactions have had aggressive undertones. Whether it is about performance, SSR, or configuration. It feels that there's a push to do things a particular way, but with few regards of the constraints surrounding it.

I don't know if this is just me, but I feel it's worth bringing up that it is stressing me out. To the point where my first reaction to this issue was: _"nope, closing this — I'm not engaging in another argument about introducing flags."_ But that doesn't seem constructive, so I'm hoping that by explaining how I'm feeling we might resolve things differently.

[...]

PR #352, opened 2017-12-07
Ignore .babelrc files of dependencies.

goto-bus-stop:
One attempt at fixing the issue described in #348: Some npm packages don't exclude `.babelrc` from the package, which causes babel to try to transform the package's files with their custom babel config. but, usually packages don't specify babel transforms as production deps, so that doesn't work.

this patch uses [tfilter](https://npmjs.com/package/tfilter) to apply different babelify options for different files. namely, files in node_modules are transformed with babel-preset-env _only_, ignoring their own .babelrc configs. local files are transformed with babel-preset-env by default but local .babelrc config is merged in so folks can add their own babel plugins too (this is the default babelify behaviour).

the babelify config for local files only also gets the `global: true` option, but that's only to make sure that the order stays the same as it was before (so it runs _after_ yoyoify etc). i think it might be best to change that in the future, so that babelify runs earlier, but it could be a breaking change so i held off on that for now.

 - [x] todo add a test for a dependency with a .babelrc file?

### DIFF
--- a/lib/graph-script.js
+++ b/lib/graph-script.js
@@ -1,6 +1,7 @@
 var debug = require('debug')('bankai.node-script')
 var concat = require('concat-stream')
 var to = require('flush-write-stream')
+var tfilter = require('tfilter')
 var assert = require('assert')
 
 var babelPresetEnv = require('babel-preset-env')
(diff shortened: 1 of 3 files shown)


toddself:
@goto-bus-stop this is working super well for me!  I'm gonna do a little more testing, but yesssssssssss

PR #355, opened 2017-12-13
Compare asset creation date on build

jondashkyle:
We should only copy assets when they are newer than what is already in `dist/`! Uses [fs-compare](https://github.com/jdeal/fs-compare).

This is especially handy when you have a Dat archive in your `dist/` directory and you don’t want to pollute the file history with new versions of assets every time you build. This is especially true if those assets are 500mb of videos, like in my case right now!

### DIFF
--- a/lib/cmd-build.js
+++ b/lib/cmd-build.js
@@ -1,6 +1,7 @@
 var ansi = require('ansi-escape-sequences')
 var pinoColada = require('pino-colada')
 var async = require('async-collection')
+var fsCompare = require('fs-compare')
 var pumpify = require('pumpify')
 var iltorb = require('iltorb')
 var mkdirp = require('mkdirp')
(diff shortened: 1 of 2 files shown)


lib/cmd-build.js
```diff
@@ -85,8 +86,13 @@ function build (entry, opts) {
         // on some OSes it is zero-copy all the way; e.g. never leaves the
         // kernel! :D
         function copy (done) {
-          if (fs.copyFile) fs.copyFile(src, dest, fin)
-          else pump(fs.createReadStream(src), fs.createWriteStream(dest), fin)
+          fsCompare.ctime(src, dest, function (err, diff) {
```
Flet: Should this be modified time instead of created time?

jondashkyle: Of course—good catch!

yoshuawuyts:
@jondashkyle dope! — you cool if we merge this? :grin:

jondashkyle:
@yoshuawuyts very down if you are!

PR #357, opened 2017-12-19
Tweak generated SSL key to make Chrome happy

quackingduck:
The version of chrome on my machine (63) requires that the
subjectAltName field of the certificate be the same as the
hostname or it won't accept the certificate regardless of
whether it is "trusted" by the system.

Also, on MacOS users can run 

```
sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain ~/.config/bankai/cert.pem
```

to add the certificate to their system keychain and then it will "just work" in all the browsers. Might be worth mentioning in the docs.

### DIFF
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -133,7 +133,18 @@ function createKeys (cb) {
 
       var opts = {
         days: 2048,
-        algorithm: 'sha256'
+        algorithm: 'sha256',
+        extensions: [
+          {
+            name: 'subjectAltName',
+            altNames: [
+              {
+                type: 2,
+                value: 'localhost'
+              }
+            ]
+          }
+        ]
       }
 
       selfsigned.generate([{ name: 'commonName', value: 'localhost' }], opts, function (err, keys) {
--- a/lib/http-server.js
+++ b/lib/http-server.js
@@ -139,7 +139,7 @@ function createKeys (cb) {
             name: 'subjectAltName',
             altNames: [
               {
-                type: 2,
+                type: 2, // DNSName
                 value: 'localhost'
               }
             ]


lib/http-server.js
```diff
@@ -133,7 +133,18 @@ function createKeys (cb) {
...
       var opts = {
         days: 2048,
-        algorithm: 'sha256'
+        algorithm: 'sha256',
+        extensions: [
+          {
+            name: 'subjectAltName',
+            altNames: [
+              {
+                type: 2,
```
goto-bus-stop: could you add a comment here about what `type: 2` means?

quackingduck: The resource I used for that is: https://github.com/digitalbazaar/forge/blob/master/lib/x509.js which has lots of great comments. Would you like me to include the url for more context?

goto-bus-stop:
Oh yeah, that looks like a great resource to link to :D


On 10 January 2018 01:41:55 CET, Myles Byrne <notifications@github.com> wrote:
>quackingduck commented on this pull request.
>
>
>
>> @@ -133,7 +133,18 @@ function createKeys (cb) {
> 
>       var opts = {
>         days: 2048,
>-        algorithm: 'sha256'
>+        algorithm: 'sha256',
>+        extensions: [
>+          {
>+            name: 'subjectAltName',
>+            altNames: [
>+              {
>+                type: 2,
>
>The resource I used for that is:
>https://github.com/digitalbazaar/forge/blob/master/lib/x509.js which
>has lots of great comments. Would you like me to include the url for
>more context?
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/choojs/bankai/pull/357#discussion_r160564115

-- 
Sent from mobile. Please excuse my brevity.

goto-bus-stop:
hella thanks! :D

goto-bus-stop:
📦 v9.2.0

Issue #358, opened 2017-12-22
WASM Brotli


yoshuawuyts:
We should investigate using this - https://github.com/dfrankland/wasm-brotli. Having to compile at the moment is slow, and a lil annoying haha.

yoshuawuyts:
Tried running this with a static server, but it seems it doesn't work as expected. Probably an issue with `wasm-brotli` 😢 

```js
var http = require('http')
var path = require('path')
var fs = require('fs')

http.createServer(function (req, res) {
  res.writeHead(200, {
    'Content-Type': 'text/html',
    'Content-Encoding': 'br'
  })

  fs.createReadStream(path.join(__dirname, 'index.html.br'))
    .pipe(res)
}).listen(8085, function () {
  console.log('listening on 8080')
})
```

dfrankland:
Yep, there definitely seems to be some issue with `compress`.

Just adding changing 1 character in a string can break it apparently. Take a look at the example below:

```js
const { compress, decompress } = require('wasm-brotli');

const badHtml = `
  <!doctype html>
  <html>
    <head>
      <title>Testing!</title>
    </head>
    <body>
      <h1>Hello, world!</h1>
    </body>
  </html>
`;

const goodHtml = `
  <!doctype html>
  <html>
    <head>
      <title>Testng!</title>
    </head>
    <body>
      <h1>Hello, world!</h1>
    </body>
  </html>
`;

(async () => {
  await Promise.all([goodHtml, badHtml].map(async (content) => {
    try {
      const buffer = Buffer.from(content, 'utf8');
      const compressed = await compress(buffer);
      const decompressed = await decompress(compressed);
      const newContent = Buffer.from(decompressed).toString('utf8');
      console.log(`This one works!\n${newContent}`);
    } catch (err) {
      console.error(`This one failed!\n${content}`);
    }
  }));
})();
```

The only character that is changed is the `i` in `Testing!`.

goto-bus-stop:
thanks for digging into this @dfrankland! for reference, looks like this was the core issue: https://github.com/dropbox/rust-brotli/issues/15

danielrh:
I fixed the bug in brotli: can you update the crate to 1.1.1 and try again?

dfrankland:
I've ran a couple of tests with `wasm-brotli@1.0.2` and everything seems to work now that it includes the `brotli@1.1.1` crate (thank you @danielrh).

yoshuawuyts:
[...]

PR #360, opened 2017-12-29
Swap iltorb for wasm-brotli

goto-bus-stop:
Ref #358 

This seeeeeems to work already!

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "through2": "^2.0.3",
     "tinyify": "^2.4.0",
     "v8-compile-cache": "^1.1.0",
-    "wasm-brotli": "^1.0.1",
+    "wasm-brotli": "^1.0.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,6 +72,7 @@
     "through2": "^2.0.3",
     "tinyify": "^2.4.0",
     "v8-compile-cache": "^1.1.0",
+    "wasm-brotli": "^1.0.1",
```
dfrankland: @goto-bus-stop can you try upgrading to `>= 1.0.2`? That will have the fix for 32-bit targeted rust crates.

yoshuawuyts: I just tried this, and it's working great! Yay :sparkles:

goto-bus-stop:
awesome, thanks a ton @dfrankland 💃  this is super cool!

Issue #337, opened 2017-11-22
Having a large amount of divs slows bundling by a lot


nicknikolov:
example code in my view after I create choo app with `create-choo-app` module:
```
  function weeks () {
    var divs = []
    for (var i = 0; i < 4000; i++) {'
      var week = html`<div class= dib w1 bg-yellow">
        ${i}
      </div>`
      divs.push(week)
    }

    return divs
  }

  return html`
    <body class="bg-washed-yellow">
      <div class="ph3 f0">
        ${weeks()}
      </div>
    </body>
  `
```
this takes several seconds to bundle, completely kills live-reload workflow.

nicknikolov:
in fact, maybe the correct thing to say is that SSR takes a lot of time not bunding:
@goto-bus-stop please correct me

goto-bus-stop:
yep we dug into this on IRC and the problem is that, when SSR generates a lot of html, the documentify pipeline takes a very long time to process it. that's because bankai adds about 10 transforms that each parse the full html file to add a few small things.

two ideas that may help fix:

 - collapsing transforms, adding all bankai's tags in a single hyperstream pass
 - inserting the SSR'd html somewhere near the end of the pipeline using a new transform, so that only `post` transforms will run on it (once that lands in documentify), and not bankai's document setup transforms.

bjrnt:
Is there anything that the end-user can currently do to alleviate this? I'm currently experiencing the same issue. Is there any way to turn of SSR for development builds?

goto-bus-stop:
One workaround is to go into node_modules/bankai/lib/graph-document.js, and replace the line `var body = res.body` with `var body = "`. Not ideal :(

goto-bus-stop:
8eab621 helps, but not enough!

nicknikolov:
I reckon we should keep this issue open as a general UX issue?

yoshuawuyts:
@nicknikolov yeah, probably (:

goto-bus-stop:
hstream helped a lot I think. of course it's always good to be faster but I think it is not terrible anymore. do you reckon this can be closed?

nicknikolov:
I've moved on from that project so can't test currently. So i'm gonna close it, if it comes up again for me i might reopen. Thanks for the help on that issue!

PR #365, opened 2018-01-03
Collapse hyperstream

goto-bus-stop:
This changes `graph-document` to only use a single `hyperstream` to add all our `<head>` stuff. This should help speed things up a lot when SSR results in a large amount of html, since we only need to parse the html once (or twice during `build`, for inlining critical css) instead of ~10 times.

### DIFF
--- a/lib/graph-document.js
+++ b/lib/graph-document.js
@@ -68,15 +68,15 @@ function node (state, createEdge) {
       var html = head(body, language)
       var d = documentify(entry, html)
       var header = [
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
snippet in #337 went from 12s response time to 3s--a big improvement, but still really slow. this needs a bit more work

yoshuawuyts:
@goto-bus-stop are you planning to merge this? Looks like a really cool patch!

PR #369, opened 2018-01-06
Browserify 15

goto-bus-stop:
Adds support for object rest spread `{ ...a }` and requiring things using template literals ```require(`a`)``` (no interpolation, so ```require(`${a}`)``` does not work)


goto-bus-stop:
just realising that many transforms don't support rest spread, so there's that

PR #370, opened 2018-01-07
Exclude electron module from browserify transform in pre 9.0.0

lejeunerenard:
If left in, electron does not properly load the `electron` module but instead tries to load the app binary directly via the `path.txt`.

I believe the correct course of action should be to exclude the `electron` module from the browserify transform. This allows electron to load the module via the built in node `require()` in electron. It seems electron does some magic for it's own module by loading from its asar.

**Note:** I'm not sure which branch I should use for pre `9.0.0` patches like this. Let me know if anything needs to change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,6 +88,10 @@ function Bankai (entry, opts) {
       ? watchify(browserify(jsOpts))
       : browserify(jsOpts)
 
+    if (opts.electron) {
+      b.exclude('electron')
+    }
+
     if (!self.cssDisabled) {
       b.plugin(cssExtract, { out: createCssStream })
       b.ignore('sheetify/insert')


yoshuawuyts:
This seems quite reasonable! - We're not really maintaining bankai pre v9 tho. Is there something in v9 that isn't working out for you?

I think if you want to continue working with the v8 branch, forking would be the right course of action. But yeah, we'd be really interested in hearing how to improve v9 so it works for you! :grin:

lejeunerenard:
Nope. There is nothing wrong with v9 per se. I just have not yet tried v9 yet because I discovered this issue when working with `create-choo-electron` which currently pins `bankai` to use v8.

I saw an issue (https://github.com/choojs/create-choo-electron/issues/8) about upgrading `create-choo-electron` to use `bankai` v9 and assumed that v9 wasn't ready for use with it. So I was thinking that this fix could just be a bug patch release under v8 to fix the problem until v9 was ready for `create-choo-electron`.

I would be happy to test out bankai v9 with `create-choo-electron`!

yoshuawuyts:
We're making progress on adding Electron support in https://github.com/choojs/bankai/pull/384 - having your thoughts would be great!

goto-bus-stop:
will close as we won't be working on v8 more in this repo. thanks!

PR #371, opened 2018-01-07
Replace purifycss by purgecss

Ffloriel:
## Proposed changes

[PurgeCss](https://github.com/FullHuman/purgecss) is a tool similar to purifycss. It is inspired by purifycss but fixes a lot of purifycss issues. One of those issues is the fact that purifycss does not take number into consideration.
e.g: `.classname1` and `.classname2` are the same for purifycss.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ can be useful to know which optimizations we apply. This is a list:
 ### CSS
 - __sheetify:__ extract all inline CSS from JavaScript, and include it in
   `bundle.js`.
-- __purifyCSS:__ removes unused CSS from the project.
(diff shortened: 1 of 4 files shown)


goto-bus-stop:
sorry, reverting this for now in #403; it was removing the `hljs-keyword`, `hljs-string` and other classes from highlight.js on the choo.io website.

goto-bus-stop:
gotcha, thanks for the explanation! i think whitelistPatterns is not going to work very well for bankai unfortunately, because we don't know what classes projects are going to be using either. we could whitelist `hljs` but someone else might use a different library that runs into the same issue.

PR #373, opened 2018-01-09
Update regex router

### LINKED ISSUE #367: Regex router ignoring some valid paths
In `bankai/http`, the regex-router ignore some valid paths like

```txt
/index.dev.js
```

The above route should be cauhgt by the js regex, but this regex only expect 1 dot, the one before the js extension.

```js
/\/([a-zA-Z0-9-_]+)\.js$/
```

So that route is handle by the default route which set a `text/html` content-type header (which will throw in some cases). This problem is extensible for css ~and source map~ files, but I don't think those will throw (still feels wrong to serve a css file like an html file)
**edit:** This doen's affect source maps, in fact source map consider the extra dots, this should be applied to js and css files too.

A second wrong route parsing is with queryStrings like

```txt
/assets/my-file.txt?version=1.0
```

The assets route will catch that, but will look for a file named `my-file.txt?version=1.0` which doesn't exists, so it will throw a 404 error. this is because the regex

```js
/^\/assets\/(.*)$/
```

just catch anything after `/assets/`, including the querystring. Not sure of the impact of this, but I guess some external services, like cdn use querystrings on files (it happened to me with workbox service).

I hope now this is clear enough to solve the issue.

YerkoPalma:
fix #367 

Not sure if we should also ignore querystrings for other files beside assets.

### DIFF
--- a/http.js
+++ b/http.js
@@ -117,7 +117,7 @@ function start (entry, opts) {
     })
   })
 
-  router.route(/\/([a-zA-Z0-9-_]+)\.js$/, function (req, res, params) {
+  router.route(/\/([a-zA-Z0-9-_.]+)\.js$/, function (req, res, params) {
     var name = params[1]
     compiler.scripts(name, function (err, node) {
       if (err) {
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
thankss! i think it makes sense to ignore query strings on JS and CSS files too. we need to keep them for SSR pages but i think everyone else can ignore them.

YerkoPalma:
@goto-bus-stop done :+1: ignored querystring in js and css files

goto-bus-stop:
super, thanks 🎉

goto-bus-stop:
📦 v9.2.0

PR #375, opened 2018-01-11
Dependecies update ...

lguzzon:
Dependencies update using next-update (no regression inside "npm test" passed) 
PS
  May be you can use it (next-update) to stay update with the latest libs ;) with a weekly CI-build using the tool to check/update the dependencies


goto-bus-stop:
oh thanks! hmm, i don't like the pinning of dependencies to a single patch version that happens here; most of these updates, we get already because of semver. i'd be cool with a PR that updates those packages that have new major versions tho!

yoshuawuyts:
Hey, looks like this patch is conflicting with `master` right now too. Seconded what @goto-bus-stop says; it'd be cool to have all the new semver majors in, but wit keeping the semver `^` identifier so we get automatic updates :D

goto-bus-stop:
@lguzzon would you like to update this PR to use ^?

package.json
```diff
@@ -16,77 +16,78 @@
...
-    "fs-compare": "0.0.4",
+    "fs-compare": "^0.0.4",
     "get-port": "^3.2.0",
     "glslify": "^6.1.0",
-    "gzip-size": "^3.0.0",
-    "hstream": "^1.0.0",
+    "gzip-size": "^4.1.0",
+    "hstream": "^1.1.0",
     "html-minifier": "^3.5.3",
+    "html-minifier": "^3.5.8",
```
goto-bus-stop: oops, the html-minifier line got duplicated somehow 🙈

lguzzon: You are right !!!

PR #384, opened 2018-01-17
Detect if a project is an electron project

yoshuawuyts:
This sets us up to add Electron support for Bankai. I was thinking that if we know a project has Electron as a dependency, we could launch it using their local Electron version. And build it using their local Electron-Builder.

Before any of that can be done, we need to detect if we're dealing with an Electron project in the first place. This does exactly that.

We chose to do add this logic to `cmd-start.js` first, because it doesn't feel like it belongs in either `index.js` or `http.js`. And not in `bin.js` either, because `lib/cmd-inspect` has no use for this either.

Hope this makes sense! Keen to receive feedback on this!

---
Paired with @blahah on this :sparkles:

### DIFF
--- a/lib/cmd-start.js
+++ b/lib/cmd-start.js
@@ -1,28 +1,35 @@
 var getPort = require('get-port')
 
-var bankai = require('../http')
+var isElectronProject = require('./is-electron-project')
 var http = require('./http-server')
+var bankai = require('../http')
 
 module.exports = start
 
 function start (entry, opts) {
-  var handler = bankai(entry, opts)
-  var state = handler.state // TODO: move all UI code into this file
-  var server = http.createServer(function (req, res) {
-    if (req.type === 'OPTIONS') return cors(req, res)
-    handler(req, res, function () {
-      res.statusCode = 404
-      return res.end('No route found for ' + req.url)
-    })
-  })
+  isElectronProject(process.cwd(), function (err, bool) {
+    if (err) throw err
+    opts.electron = bool
 
-  getPort({port: 8080}).then(function (port) {
-    server.listen(port, function () {
-      state.port = port
+    var handler = bankai(entry, opts)
+    var state = handler.state // TODO: move all UI code into this file
+    var server = http.createServer(function (req, res) {
+      if (req.type === 'OPTIONS') return cors(req, res)
+      handler(req, res, function () {
+        res.statusCode = 404
+        return res.end('No route found for ' + req.url)
+      })
     })
-  })
-  .catch(function (err) {
-    state.error = err
+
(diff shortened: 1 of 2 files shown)


lib/cmd-start.js
```diff
@@ -1,28 +1,35 @@
...
-  var handler = bankai(entry, opts)
-  var state = handler.state // TODO: move all UI code into this file
-  var server = http.createServer(function (req, res) {
-    if (req.type === 'OPTIONS') return cors(req, res)
-    handler(req, res, function () {
-      res.statusCode = 404
-      return res.end('No route found for ' + req.url)
-    })
-  })
+  isElectronProject(__dirname, function (err, bool) {
```
goto-bus-stop: wouldn't this always end up checking `bankai`'s own `package.json`? 🤔

yoshuawuyts: oh lol, totally right!

YerkoPalma: `process.cwd()` would be a best fit?

yoshuawuyts: fixed!

goto-bus-stop: yeah, `process.cwd()` or `path.dirname(entry)` are both reasonable I think

lejeunerenard:
**Detecting Electron**  
I like detecting if the current project is running `electron` via the dependencies. This don't cover the unlikely situation where there is no `package.json` or where `electron` is not a dependency. But the alternative is parsing files and checking for any that require `electron`, which would be gross.

These edge cases could be supported by providing an `--electron` option like v8 had so that users could manually override electron support. This would also allow users to disable the electron specific work if necessary. Then again... a flag goes against the configuration philosophy of bankai. 😄

> The Bankai CLI doesn't take any flags, other than to manipulate how we log to the console.

**Where to place the logic**  
I agree that placing electron detection in `index.js`, `http.js` & `bin.js` doesn't make sense. But doesn't `lib/cmd-build.js` need the `opts.electron` value as well to build out the assets with electron specific configurations ( such as excluding the `electron` package from browserify, etc )?

I'm excited for electron support in v9! ヽ(•‿•)ノ

yoshuawuyts:
@lejeunerenard haha, yeah seems we're having the same considerations. `lib/cmd-build.js` would also need this logic, but in turn it means that `lib/cmd-inspect.js` could be kept fast. Not the worst haha.

Also think we can deffo get this into the `8.x` release line! :D No breaking changes necessary afaict :sparkles:

yoshuawuyts:
> seems to work! were you hoping to get this in already and then do the electron starting stuff later in a separate pr?

yeah! - figured small patches are easier to review & rebase! :D

Issue #382, opened 2018-01-16
v9.2.0 not loading static assets


gavinpatkinson:
I updated bankai but it no longer loads static assets from ./assets. Without changing anything but the version no assets are loaded.

yoshuawuyts:
@gavinpatkinson oh, that's bizarre. Shouldn't be happening! Are you seeing any errors anywhere?

gavinpatkinson:
No error, bankai bundles and serves. No error in console either.  I was wondering if the src path that the app expects has changed – right now I'm using something like `const imageURL = 'assets/images/' + project.filename`

yoshuawuyts:
@gavinpatkinson I'm having trouble reproducing this, https://github.com/yoshuawuyts/repro-bankai-382/blob/master/index.js#L18-L27 seems to be working as expected.

```js
app.use((state, emitter) => {
  window.fetch('/assets/icon.png')
    .then((res) => {
      console.log('ok', res.ok)
      console.log(res)
    })
    .catch((err) => {
      emitter.emit('log:error', err)
    })
})
```

![2018-01-18-014336_1920x1080](https://user-images.githubusercontent.com/2467194/35075052-655a765c-fbf1-11e7-9fbb-6ef8f313f535.png)


Could you perhaps try cloning https://github.com/yoshuawuyts/repro-bankai-382, and give it a shot. If you find a failure scenario, a PR to the repo might be useful.

Thanks, and sorry this is being a bit of a pain to track down. I'm confident we'll be able to get to the bottom of it tho! Thanks for your patience :grin:

greggraf:
I have observed a similar change in behavior. 

Whether it is an appropriate use of the `/assets/` directory or not, previously I was able to get `/assets/file.json` to serve from there.  

I am able to reproduce the use case @gavinpatkinson has (image files in a subdirectory of `/assets/` ie: `/assets/images/image.jpg`) and it seems to be failing for a similar reason. 

In https://github.com/choojs/bankai/commit/79d2aad6b9a135f972265e6edf9019524800aa49 there are some RegExs that change previous behavior.

The one for assets is affecting @gavinpatkinson.

Previous RegEx behavior for assets:
```
> '/assets/image.png'.search(/^\/assets\/(.*)$/)
< 0
> '/assets/images/image.jpg'.search(/^\/assets\/(.*)$/)
< 0
```
Current RegEx behavior for assets:
```
[...]

PR #396, opened 2018-01-22
Restore asset serving behavior

greggraf:
Attempts to address #382 

I added some tests around routing and updated the regex's for scripts, styles and assets.
Additionally I moved the asset route above the style and script routes so .js and .css files can be served from the assets/ directory without being first matched by the script and style compiler routes.

### DIFF
--- a/http.js
+++ b/http.js
@@ -117,7 +117,7 @@ function start (entry, opts) {
     })
   })
 
-  router.route(/\/([a-zA-Z0-9-_.]+)\.js\?{0,1}[\w&=]*$/, function (req, res, params) {
+  router.route(/\/([a-zA-Z0-9-_.]+)\.js(\?.*)?$/, function (req, res, params) {
     var name = params[1]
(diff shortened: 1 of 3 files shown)


http.js
```diff
@@ -117,7 +117,19 @@ function start (entry, opts) {
     })
   })
 
-  router.route(/\/([a-zA-Z0-9-_.]+)\.js\?{0,1}[\w&=]*$/, function (req, res, params) {
+  router.route(/^\/assets\/([^?]*)(\?.*)?$/, function (req, res, params) {
```
yoshuawuyts: Dang yeah, we can't quite merge this without also supporting `content` - it'd otherwise cause a regression. Probably can be done a s a quick hack tho :sparkles:

yoshuawuyts: Oh lol, no so this is just us. LGTM!

yoshuawuyts:
`v9.4.0`

PR #397, opened 2018-01-23
Support assets in nested folders

emilbayes:
This expands all RegEx's to accept / as a valid char in the path matching group

### DIFF
--- a/http.js
+++ b/http.js
@@ -117,7 +117,7 @@ function start (entry, opts) {
     })
   })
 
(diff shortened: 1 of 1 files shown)


YerkoPalma:
This seems to be highly related to #396. I think both PR should be merged or something

emilbayes:
I actually think #396 makes this PR obsolete. That PR also has tests and is more thorougher, just didn't see it before I submitted this

YerkoPalma:
:open_mouth: you are right, didn't see that included tests have nested folders in it.

yoshuawuyts:
Cool, yeah closing in favor of #396 then! :sparkles:

Issue #309, opened 2017-10-24
Allow setting the base path for the application


Flet:
Currently, `graph-document.js` sets all URLs to bundles as `/`, which means the application must be deployed to the root of a site in order to work.

Allowing a user to override this would ensure applications can be deployed to a subdirectory on a domain without issue (i.e. http://example.com/mychooapp/is/way/down/here).

Unfortunately there is not a way to get around this without modifying the generated HTML files after `bankai` runs.

finnp:
Also routers are confused because of this. In development mode they should be able to server `/` from `localhost`, but in production it could be `/berlin` as root for example.

Furthermore when the static routes are created, it should ignore the "root path".

I think a solution could be to somehow pass `{"root": "/berlin"}` to bankai.

Then when building dist it will will NOT put files in a subdirectory `/berlin`, but the referenced paths are prefixed with `/berlin`.

Also it would set `process.env.ROOT_PATH` to `/berlin`, so it can be passed to routers in the frontend (for normal `npm start` this would be `''`).

goto-bus-stop:
#399

PR #399, opened 2018-01-25
Add baseHref option

emilbayes:
I'm not sure how this PR sits with the philosophy of bankai being almost zero-conf. I'm also not sure that this fills out all the places that a `<base href="some-url" />` would cover. Lets' discuss!

Ref: #309

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -58,7 +58,7 @@ var argv = minimist(process.argv.slice(2), {
     help: 'h',
     quiet: 'q',
     version: 'v',
-    'base-href': 'baseHref'
+    'base': 'base'
   },
   boolean: [
     'help',
--- a/bin.js
+++ b/bin.js
@@ -57,7 +57,8 @@ var argv = minimist(process.argv.slice(2), {
   alias: {
     help: 'h',
     quiet: 'q',
-    version: 'v'
+    version: 'v',
+    'base-href': 'baseHref'
   },
   boolean: [
     'help',
--- a/lib/cmd-build.js
+++ b/lib/cmd-build.js
@@ -17,7 +17,8 @@ module.exports = build
 function build (entry, outdir, opts) {
   var bankaiOpts = {
     logStream: pumpify(pinoColada(), process.stdout),
-    watch: false
+    watch: false,
+    baseHref: opts.baseHref
   }
   var compiler = bankai(entry, bankaiOpts)
   var log = compiler.log
--- a/lib/cmd-build.js
+++ b/lib/cmd-build.js
@@ -18,7 +18,7 @@ function build (entry, outdir, opts) {
   var bankaiOpts = {
     logStream: pumpify(pinoColada(), process.stdout),
     watch: false,
-    baseHref: opts.baseHref
+    base: opts.base
   }
   var compiler = bankai(entry, bankaiOpts)
   var log = compiler.log
--- a/lib/graph-document.js
+++ b/lib/graph-document.js
@@ -64,7 +64,7 @@ function node (state, createEdge) {
       var language = res.language
       var title = res.title
       var body = res.body
-      var baseHref = state.metadata.opts.baseHref
+      var base = state.metadata.opts.base
 
       var hasDynamicScripts = state.scripts.list.buffer.length > 0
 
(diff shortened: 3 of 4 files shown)


YerkoPalma:
I guess the base directory could be guessed, so we can keep bankai philosophy. For instance,

```bash
$ bankai start app/index
```

should result in a basehref of `app`. Something like this should be enough:

```js
var baseHref = path.dirname(entry)
```

goto-bus-stop:
I'd really love if you could drop a bankai built app into any directory and have it just work, without hardcoding the base path…

I thought we could maybe use relative URLs for scripts. when bankai renders routes it knows if it's on `/` or `/deeply/nested/thing` and could make relative paths accordingly. but you can also make all requests hit index.html for a non server rendered app and then the paths would be wrong, so I guess that's not an option. might be a good area for more exploration though.

(it looks like the client side router doesn't work when deployed to a subdirectory right now either, if you do not use hash routing)

Is there a reason why we can't use `<base href="${opts.baseHref}" />`? of course it would be better to have an option-less thing, but `<base href>` seems easier than adding it manually in all places?

Adding `<base href="someroot">` could be done already by installing `hstream` and adding a documentify transform to package.json:

```json
{
  "documentify": {
    "transform": [
      ["hstream", { "head": { "_appendHtml": "<base href='someroot'>" } }]
    ]
  }
}
```

of course that's not ideal but it might be helpful as a noninvasive way to try out the `base href` way.

@YerkoPalma would that address the case where you have a standalone bankai app but deploy it to `xyz.com/app/` (or eg `tilde.town/~username/`)? If you would just start it as `bankai start index.js` locally in its own git repo.

emilbayes:
Adding a `<base />` tag would be too much magic IMO, as that would impact everywhere that resources are loaded with a leading slash

YerkoPalma:
@goto-bus-stop it works for those and I guess that almost every example, at least in unix systems

image

emilbayes:
Instead of having this strand, I think @YerkoPalma proposal was good and have the default base HREF `.`

bin.js
```diff
@@ -81,6 +82,7 @@ var argv = minimist(process.argv.slice(2), {
   } else if (argv.version) {
     console.log(require('./package.json').version)
   } else if (cmd === 'build') {
+    console.log(argv)
```
yoshuawuyts: ^

goto-bus-stop:
> Adding a `<base />` tag would be too much magic IMO, as that would impact everywhere that resources are loaded with a leading slash, including dependencies you may not control

@emilbayes that sounds like a really rare case to me for web dependencies, but maybe it is different in electron? I'm not very familiar with it honestly.

I would really rather not introduce new config. https://github.com/choojs/bankai/pull/384 added electron project detection—maybe we can use it to do some magic that works for electron projects as a first step?

yoshuawuyts:
Think this approach is pretty neat. Perhaps we could rename `--base-href` to `--base` in the CLI for ergonomics reasons?

yoshuawuyts:
@goto-bus-stop there was also the case of deploying on subdomains that's related to this:

> Allowing a user to override this would ensure applications can be deployed to a subdirectory on a domain without issue (i.e. http://example.com/mychooapp/is/way/down/here).

Think that's a valid usecase for a lot of people - deployment envs can often be out of control of local devs, and less than ideal. Adding an opt would def solve multiple issues at once for folks heh.

goto-bus-stop:
yeah, i guess we don't have many other options here. @emilbayes I think this approach is good, if you rename it to --base we can get this in.

@yoshuawuyts i believe nanorouter has a similar issue and does not work correctly inside a subdirectory unless you use hash routing only; maybe we should also add a `base` or `root` property to it? I'll check if there's an issue open about that already

emilbayes:
@goto-bus-stop Do you also want the internal option and variable names to be just `base`?

goto-bus-stop:
i think it would make the most sense for them to be the same, yes

yoshuawuyts:
updated to use `'base'`

ZhouHansen:
@emilbayes There's an issue, run:
```
 bankai build index --base boowa
```
generate: 
```
<script src='boowa,boowa/ac78130558410d00/bundle.js'>
```

PR #403, opened 2018-01-27
Revert "Replace purifycss by purgecss (#371)"

goto-bus-stop:
This reverts commit 69f0ac80ebee76f774250cef6cc9a12463c7a4e3.

purgecss was removing the keyword classes from `hljs`, breaking the
highlighting on the choo website. revert to purify-css for now, which is
a bit more conservative.


goto-bus-stop:
📦 9.4.1

PR #406, opened 2018-01-31
Add TypeScript support.

goto-bus-stop:
If `typescript` is installed, add the [tsify](https://github.com/typestrong/tsify) plugin.

If a syntax error occurs in a `.ts` file, recommend installing
`typescript`.

image

TODO: Catch and format errors from the typescript compiler…

image


lib/is-electron-project.js
```diff
@@ -1,24 +0,0 @@
-var explain = require('explain-error')
-var findup = require('findup')
-var path = require('path')
-var fs = require('fs')
-
-module.exports = isElectronProject
-
-function isElectronProject (dirname, cb) {
```
goto-bus-stop: Moved this to a `hasDependency` util function because the code needed to check for typescript was exactly the same.

goto-bus-stop:
Now also formatting typescript errors:

image

And did some work on syntax error messages for javascript too:
image
(Previously this would only show "Syntax Error")

yoshuawuyts:
@goto-bus-stop sick work! Question tho: none of this would work with server rendering, right?

goto-bus-stop:
Afraid that's correct 😄 that is also going to be an issue for react support with babel. Not sure what to do about that yet.

Issue #410, opened 2018-02-05
feature request: user provided state in bankai SSR


cdaringe:
# problem

- i want to embed bankai in my server per #398, but need the rendered route to be derived via external state (e.g. my user's state)

# discussion

- existing discussion [here](https://github.com/choojs/bankai/blob/3057e9e4e9ee61cd5f5e1c4149926c0577635000/lib/graph-document.js#L58)
  - maybe a `e.on('prerender', (route, state, cb) => ())` API would fill the need?
- i also need a convention for hydrating that base state browser-side as well.  maybe we can collab on this in tandem here?

🚞🚞🚞

goto-bus-stop:
choo reads state from the `window.initialState` variable, so we can output code for that in bankai.

cdaringe:
neat-o. let's move the convo over to #413!

PR #413, opened 2018-02-06
feat(ssr): optionally render with app state

cdaringe:
# problem

- see https://github.com/choojs/bankai/issues/410

# solution

- prototype api to allow user to provide app state to render with

# discussion

im not married to this implementation at all, but i am eager for the feature, so i figured some code would get the convo rolling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ npm-debug.log*
 .DS_Store
 dist/
 *.swp
+yarn.lock
+.vscode
--- a/README.md
+++ b/README.md
@@ -246,8 +246,9 @@ var compiler = bankai(
   ...,
   {
     ...,
-    onPreRender: function (cb) {
-      cb(null, { animal: 'cat' }) // callback with application state
+    onPreRender: function (route, cb) {
+      var initialState = { animal: 'cat' }
+      cb(null, initalState) // callback with application state
     }
   }
 )
--- a/README.md
+++ b/README.md
@@ -237,6 +237,22 @@ server.listen(8080, function () {
 })
 ```
 
+## Server Side Rendering (SSR)
+By default, bankai will render choo routes server side if it can.  If you need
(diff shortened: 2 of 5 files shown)


cdaringe:
hi amigos, any time to poke at this yet?  i really want to emphasize again that i'm ok with this PR being  throw-away-ready if teamsters have different ideas.  im just eager to get the feature in or on the roadmap!  thx

goto-bus-stop:
Aye, thanks for the PR!

My first thought is that we probably need to know the route in an `onPreRender` like hook; so you can initialise different state for different pages.

IIRC the `documentNode` is only initialised once, so the hook would only be called the first time. Maybe we can have an API like this somehow?

```js
var compiler = bankai(entry, opts)
compiler.documents('index.html', { initial: 'state' }, function () {})
```

Something else to consider is that we'd like to have async rendering in choo, so that you can do data loading inside the app before doing the server render. Work on that hasn't started yet but I think that might address some of the use cases for this feature too, like loading user data into state depending on the route. Doesn't mean we can't have both, especially if the choo feature is going to take some time

cdaringe:
>  we probably need to know the route

agreed. 👍 

> `compiler.documents('index.html', { initial: 'state' }, function () {})`

ive been been hacking around on this for a bit now.  im still not sure 100% how to best link all of this up.

it's as though we want each route that lands in the graph to also have an accompanying `initial-state` member in the graph as well.  if we were to use `compiler.documents(...)` API you suggest, we may have to find a way to tap into the graph and change that `initial-state` Buffer.  additionally, we would need to do something similar with a hook in the compiler options for the user to update that value, then render from the graph data in `graph-document`.  it seems like if bankai wants to render from the graph, we should go full graph, and not have a mixed mode of getting input into the render pipeline, which was what i my PR was originally doing.

initial-state ➡️  graph.  that sound right?

im a little stuck.  maybe a 🚶 will help :)

Issue #343, opened 2017-11-23
Consider support for custom index.html files?


johanalkstal:
As I understand it, bankai brings its own index.html file.

I see issues with not being able to supply your own file or template for an index.html file.
Some of the default attributes potentially block people from using bankai due to other cultural needs, like `lang` being set to american english by default, or `dir` being set to `ltr` by default.

Then of course there's the very likely need to have custom meta tags in the head of the document.

I've come to understand that there's a "hack-ish(?)" way to add things like that through `documentify`?

But I feel things like these should be as simple as possible to do without any special knowledge.

What are your thoughts?

dolebas:
In regards of custom head, what is your opinion on setting the head dynamically and redirect bots to a pre-rendering service?

johanalkstal:
@dolebas From the top of my head that sounds more complicated than things should be? But I've never had the need or attempted something similar.

dolebas:
In my experience, this approach can actually make it easier to customise the head. If you are using Netlify, they support this out of the box but other solutions like redirecting with Firebase Functions is quite straight forward as well.

johanalkstal:
In regards to the head, relying on a third party to provide you with a service doesn't sound like a reasonable default.

But dynamics is not the issue I'm talking about. It's about being in control of your document. 

The most simple solution for the user of bankai has to be to provide bankai with a index.html template. The developer has to have control over how their document looks and is structured.

dolebas:
In case you are not hosting your site on your own servers, there's actually no need to rely on an additional third party for this approach.

As your document lives in the DOM client side, your control over the document should increase when you define it with JS. Server side, as all the bots don't execute JS, a headless browser setup can provide a simple and generic solution.

I like that bankai simply takes care of the html per default but having the ability to optionally provide html templates sounds great.

goto-bus-stop:
the assumptions bankai makes about the document during SSR are also going to be a problem for adding React/Preact support. in bankai, the app is expected to handle the entire `<body>` element, but in (p)react apps you'd usually mount the app into an element like `<div id="app">`. that'd also be a useful case for custom templates.

def think we should do this!

YerkoPalma:
Definitly +1 on this because 1) Was prebiously supported by bankai, and users (like me) would expect this to keep working. 2) In choo, you also could mount your app in some other node different than body

alexbihary:
[...]

PR #414, opened 2018-02-07
Inject server rendered app at mount point.

goto-bus-stop:
Server rendering can reply with a `.selector` property indicating where
the rendered content should be mounted. It's optional; by default we
assume that we're rendering into the `<body>`.

This injects the server rendered content into the html _after_ all the
head tags are added, which should speed it up a bit for views with lots
of html. It detaches the initial template generation (the `head`
function) from injecting the server rendered response, so we can
implement custom index.html files (#343) later too.

### DIFF
--- a/lib/graph-document.js
+++ b/lib/graph-document.js
@@ -74,10 +74,11 @@ function node (state, createEdge) {
     var route = content.route
     var title = content.title
     var body = content.body
+    var selector = content.selector
 
     var hasDynamicScripts = state.scripts.list.buffer.length > 0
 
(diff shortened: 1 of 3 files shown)


cdaringe:
this looks good!  i feel like a test is warranted for the feature.

goto-bus-stop:
agree—probably need to work on the custom html feature first, because only the default case (`body`) is supported and tested here.

goto-bus-stop:
Adding test in a followup PR that also adds support for custom HTML templates. thanks!

PR #417, opened 2018-02-09
Move UI code out of http.js

goto-bus-stop:
Moves all the UI `state` stuff into lib/ui.js. the live reload connection counter `state.sse` is still managed by http.js, it emits `sse-connect` and `sse-disconnect` when the value updates so the UI can rerender. Maybe the counting should also happen in the UI code, with http.js only emitting those events?

Also added in @Flet's simple log output UI, just to make sure that the UI is indeed separate from the HTTP server now :)


Flet:
Nice thanks!

PR #419, opened 2018-02-12
Fix documentation of error event

aknuds1:
Fix the signature of the callback for the error event.


goto-bus-stop:
good catch, thanks 🎉

PR #401, opened 2018-01-27
document entry level opts

yoshuawuyts:
A few people pointed out that some of the docs weren't as clear as they could be. Hopefully this helps a little :sparkles:

PR #421, opened 2018-02-13
Fix signatures of compiler.scripts and compiler.styles

aknuds1:
This is a bug fix

## Checklist


goto-bus-stop:
thanks! this was also part of #401

PR #422, opened 2018-02-13
Add option babelifyDeps

aknuds1:
**Choose one:** This is a 🙋 feature

I add the option `babelifyDeps` to control whether the scripts method applies the babelify transform to dependencies or not. I need this since calling Babel on my dependency tree slows down my build by about 5-6 times and at least one of my dependencies (underscore.js) also breaks from this.

**Note:** I'm still unsure whether `babelifyDeps` should be true by default, since it's an intrusive feature that easily breaks or simply slows down your build significantly. Personally I would have it off by default.

## Checklist

- [x] tests pass

## Context
See #423.

## Semver Changes
I would recommend a minor version upgrade, since this adds an option.


goto-bus-stop:
IMO we should remove it entirely in 10.0 … or figure out a way to make it only compile stuff that is not ES5 compatible (maybe with https://www.npmjs.com/package/is-es-version and aggressive caching), or do a dual bundle thing—one with modern syntax and one with ES5.

In the mean time this is probably a reasonable compromise even though it adds an option.

goto-bus-stop:
will merge both this and #455 soon then and release as minor

PR #427, opened 2018-02-14
Async server render

yoshuawuyts:
- split out rendering to a class
- make logic more explicit
- support a double pass async render in choo

## Checklist

- [x] tests pass

## Context
Related to @diffcunha's experiment.

## Semver Changes
Minor

### DIFF
--- a/lib/graph-document.js
+++ b/lib/graph-document.js
@@ -1,3 +1,4 @@
+var waterfall = require('async-collection/waterfall')
 var mapLimit = require('async-collection/map-limit')
 var explain = require('explain-error')
 var concat = require('concat-stream')
@@ -9,9 +10,9 @@ var critical = require('inline-critical-css')
 var documentify = require('documentify')
 var hyperstream = require('hstream')
 
+var ttyError = require('./tty-error')
 var serverRender = require('../ssr')
 var utils = require('./utils')
-var ttyError = require('./tty-error')
 
 var WRITE_CONCURRENCY = 3
 
@@ -29,7 +30,8 @@ function node (state, createEdge) {
       return self.emit('error', 'documents', entry, err)
     }
 
-    self.emit('ssr', {success: true, renderRoute: renderRoute})
+    // TODO: don't pass a callback here - makes for messy interactions
+    self.emit('ssr', {success: true, renderRoute: documentifyRoute})
 
     var fonts = extractFonts(state.assets)
     var list = render.list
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
lol, all tests pass. Def don't merge yet tho, because we have a race condition

yoshuawuyts:
@goto-bus-stop @diffcunha if you could take this for a spin, that'd be very helpful. There's a race condition in here somewhere, and I'm staring blindly at it - I think we just need a second pair of eyes on this to figure it out. Thanks!

ssr/choo.js
```diff
@@ -0,0 +1,81 @@
...
+//
+//   update () {
+//     return true
+//   }
+//
+//   async fetch () {
+//     this.local.loaded = true // async functions return promises
+//   }
+// }
+// ```
```
yoshuawuyts: These are docs for the feature we're introducing to Choo btw

lib/graph-document.js
```diff
@@ -21,92 +22,100 @@ function node (state, createEdge) {
...
-      function complete (buf) { done(null, buf) }
-
-      pump(d.bundle(), concat({ encoding: 'buffer' }, complete), function (err) {
-        if (err) return done(explain(err, 'Error in documentify while operating on ' + route))
-      })
+    function pushEdge (buf, done) {
+      var name = route
+      if (name === '/') name = 'index'
+      name = name + '.html'
+      createEdge(name, buf)
```
yoshuawuyts: we're getting as far as having the correct buffer shop up here by the way. But then when calling the `./http.js` methods it just sends down an empty body. Looks like it never calls back or something; not  sure what's up with that.

ssr/choo.js
```diff
@@ -0,0 +1,81 @@
...
+//
+// NOTE: state is never passed in it seems. Funky fn signature, this should be
+// fixed.
+//
+// NOTE: Choo currently doesn't use the passed-in state as the base. It probably
+// should, so we can treat the state as the only stateful bits, and the rest is
+// just context.
```
goto-bus-stop: minor optimization nit, but we could not do the second render if the `_experimental_prefetch` array is empty

lib/graph-document.js
goto-bus-stop: yep you were right, it works if i pass the buffer back to the `ssr.renderRoute` call in the `done` callback one line below here

goto-bus-stop:
this is hella cool, great work :D

lib/graph-document.js
```diff
@@ -21,92 +22,100 @@ function node (state, createEdge) {
...
-    }
-
-    self.emit('ssr', {success: true, renderRoute: renderRoute})
+  var ssr = new ServerRender(entry)
+  if (ssr.error && ssr.error.isSsr) {
+    self.emit('ssr', { success: false, error: ssr.error })
+  } else if (ssr.error) {
+    ssr.error = ttyError('documents', entry, ssr.error)
+    return self.emit('error', 'documents', entry, ssr.error)
+  }
```
goto-bus-stop: are these error checks still doing anything at this point? since they are run sync after instantiating the ServerRender class

lib/graph-document.js
yoshuawuyts: nice!

lib/graph-document.js
yoshuawuyts: uhhhhhh, I don't know - probably not hey

goto-bus-stop: lets get rid of it & then merge if you think it's ready?

goto-bus-stop: oh, the `.error` property actually is sync there. nvm!

yoshuawuyts:
rebased! - should be good to merge then :D

goto-bus-stop:
Aye; I had some issues with the SSR indicator in the UI on this branch but will merge anyway so we won't have to do much rebasing when new PRs are opened. Maybe I can fix it on master, who knows!

tornqvist:
I've been experimenting with this new API, which at first felt magical, but I then encountered an issue with saving data to state for use in the second render pass.

I take it that the pattern of checking for data on state before rendering a view and emitting an event if something needs to be fetched is pretty common on choo (?)
Here's an example: https://gist.github.com/tornqvist/d7e65cfe97718f19d1aa01f0877795e7

The problem when used with this new API is that the stores reference to `state` is outdated as `toString` on choo doesn't mutate the state, but replaces it.

Does this use case make sense or have I misinterpreted the intended use of the prefetch api?

PR #429, opened 2018-02-19
use minimist correctly

ZhouHansen:
I was developing a blog generator. And I had to hard-code the src of `<script>` before deploy the blog, so I found [#399](https://github.com/choojs/bankai/pull/399) which is going to solve it, but there's a problem, when run `bankai build index.js --base=root`, the `argv` was:
```
{ _: [ 'build', 'index.js' ],
  help: false,
  h: false,
  quiet: false,
  q: false,
  version: false,
  v: false,
  base: [ 'root', 'root' ] }
```
then it will result: `<script src='root,root/ae23jf8adofk/bundle.js'></script>`,  `argv.base` should be `'root'`, so this pr fixed it


goto-bus-stop:
thanks! btw, your master branch is a bit funny, so github shows a whole ton of older and merge commits in the PR; you should be able to fix it by doing `git pull choojs master --rebase`. i think that will automatically clear up all the merge commits too. and then you can use feature branches in the future :)

goto-bus-stop:
:package: 9.7.1

Issue #435, opened 2018-02-23
Add package.json fields configuration to usage in cli


kareniel:
Configuring Bankai is done by modifying package.json. It'd be cool if you could get the usage info for the browserify, sheetify, and documentify fields in the usage text (--help) in the CLI.

#### Possible approach

Not sure about the wording, but:

1. Add this line to usage:
```
For configuration usage, type `bankai --help config`.
```

2. Output something like this:
```
Configuration


Bankai is built on top of compilers for scripts, styles and documents. 
Each of them can be configured by adding a field to your project's package.json.

These three fields are, respectively: "browserify", "sheetify" and "documentify".
Each one should have a configuration object as it's value.

There is currently one possible configuration field: "transform".

It can be one of either:

1. An array of transform names. 
       ie:  [ "babelify" ]
2. An array of tuples transform name + configuration object. 
       ie: [[ "babelify", { "presets": [ "env" ] } ]]

Full example:

{
  "browserify": {
    "transform": [
      [ "babelify", { "presets": [ "env" ] } ]
    ]
  },
  "sheetify": {
    "transform": [
      "sheetify-cssnext"
    ]
  },
  "documentify": {
    "transform": [
      [ "posthtmlify", { "use": [ "posthtml-custom-elements" ] } ]
    ]
  }
}

```

3. Profit

yoshuawuyts:
@kareniel that's a great idea!

kareniel:
Closed by #437 :)

PR #437, opened 2018-02-23
add configuration usage to cli

kareniel:
This PR adds the enhancement discussed in #435

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -36,6 +36,8 @@ var USAGE = `
     Compile all files in the project to disk
     ${clr('$ bankai build index.js', 'cyan')}
 
+  For configuration usage, type 'bankai --help config'
+
   Running into trouble? Feel free to file an issue:
   ${clr('https://github.com/choojs/bankai/issues/new', 'cyan')}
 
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
:heart_eyes:

bin.js
```diff
@@ -53,6 +55,49 @@ var NOCOMMAND = `
...
+  2. An array of tuples transform name + configuration object.
+       ie: ${clr('[[ "babelify", { "presets": [ "env" ] } ]]', 'cyan')}
+
+
+  Full example:
+
+  ${clr(`{
+    "browserify": {
+      "transform": [
+        [ "babelify", { "presets": [ "env" ] } ]
```
goto-bus-stop: I think we should use a different example, since babelify is already included with this config by default. Not sure which transform would be a good alternative tho…

kareniel: Good point! Maybe `browserify-istanbul` ? 
It's hard, all my favorite are already used by Bankai 😂

goto-bus-stop: Yeaap lol. Not sure about istanbul since it's used in production builds too. Maybe something like [vueify](https://github.com/vuejs/vueify)?

kareniel: Okay! Made the changes using vueify. The appveyor build failed though, but it seems to be unrelated.

bin.js
```diff
@@ -53,6 +55,49 @@ var NOCOMMAND = `
...
+  2. An array of tuples transform name + configuration object.
+       ie: ${clr('[[ "vueify", { "sass": { "includePaths": [ "src/assets/css" ] } } ]]', 'cyan')}
+
+
+  Full example:
+
+  ${clr(`{
+    "browserify": {
+      "transform": [
+        [[ "vueify", { "sass": { "includePaths": [ "src/assets/css" ] } } ]]
```
goto-bus-stop: i think there's one too many pairs of [] here?

goto-bus-stop: nvm, I'll just remove them while merging :)

goto-bus-stop:
yay, thanks :D yeah, appveyor fails on the branch because config-usage doesn't have the config file (it was added after you branched off) so it tries to build it as c# or whatever.

goto-bus-stop:
:package: 9.9.0

kareniel:
:D

PR #438, opened 2018-02-25
Lookup browserslist config

tornqvist:
This is a 🙋 feature

Though browsers may be configured using `.babelrc` there's no way of sharing a browserslist config with e.g. autoprefixer or cssnext without duplicating the list. This makes it easier to consolidate which browsers to support.

The tests aren't very future proof as they'd yield false positives once all major browsers catch up with the specific features used in the test.

## Checklist
- [x] tests pass

## Semver Changes
Minor


goto-bus-stop:
Sweeeet :star2:

PR #443, opened 2018-03-07
Expose mime types on graph nodes, fixes #402

### LINKED ISSUE #402: expose mime types in core API
Right now our core API only outputs the `data.buffer` and `data.hash` properties. It'd be real neat if we also exposed `data.mime` or something similar.

Turns out our `./http.js` API isn't flexible enough for e.g. `fastify`, so making it easier for them to integrate would be :100:

goto-bus-stop:
This depends on https://github.com/yoshuawuyts/buffer-graph/pull/11

Graph nodes now get a `.mime` property in addition to `.buffer` and `.hash`. Fastify can use this to set the content-type header.

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,7 @@ function Bankai (entry, opts) {
 
   // Insert nodes into the graph.
   this.graph.node('assets', assetsNode)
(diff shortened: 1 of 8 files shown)


yoshuawuyts:
https://github.com/yoshuawuyts/buffer-graph/pull/11 was merged!

goto-bus-stop:
@yoshuawuyts :raised_hands: LGTY?

PR #444, opened 2018-03-08
Fix assets list assigned an empty buffer

tornqvist:
This is a 🐛 bug fix

I noticed the font preload wasn't working because `Buffer.from(<array>)` would yield an empty buffer.

## Checklist
- [x] tests pass

## Semver Changes
Patch

### DIFF
--- a/lib/graph-assets.js
+++ b/lib/graph-assets.js
@@ -36,7 +36,7 @@ function node (state, createEdge) {
 
   tracker.on('change', function () {
     var list = tracker.list()
-    createEdge('list', Buffer.from(list))
+    createEdge('list', Buffer.from(list.join()))
   })
 
   tracker.on('progress', function (progress) {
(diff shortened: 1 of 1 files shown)


lib/graph-assets.js
```diff
@@ -36,7 +36,7 @@ function node (state, createEdge) {
 
   tracker.on('change', function () {
     var list = tracker.list()
-    createEdge('list', Buffer.from(list))
+    createEdge('list', Buffer.from(list.join()))
```
goto-bus-stop: really minor but would you mind adding the ',' argument here? i know it's the default, but we explicitly pass ',' elsewhere too!

goto-bus-stop:
thanks! nice example of a hacky implementation ending up buggy too :P

goto-bus-stop:
thanksss

PR #445, opened 2018-03-08
Exclude all choo routes with partials

tornqvist:
This is a 🐛 bug fix

Only top level partial routes were being filtered out. This excludes all routes with a partial in them.

## Checklist
- [x] tests pass

## Semver Changes
Patch


goto-bus-stop:
Thanks!

PR #446, opened 2018-03-08
Expose initial state in head

tornqvist:
This a 🙋 feature

With the support for Choo SSR using the `_experimental_prefetch` it would make sense to expose the generated initial state on the window for the client to pick up.

Idk if this is too choo specific to include in the default document but from what I could find no other framework has an official specific key under which to expose the initial state.

## Checklist
- [x] tests pass

## Context
This is in part dependent on https://github.com/choojs/choo/pull/639 which ensures a clean state on every call to `toString`.

## Semver Changes
Minor


yoshuawuyts:
@tornqvist didn't you mean to reference https://github.com/choojs/choo/pull/638 instead of the components PR. If so, that's been merged and published already!

tornqvist:
@yoshuawuyts naw, https://github.com/choojs/choo/pull/638 only added lazy stores but never forwarded the new state when reinitializing them on `toString`, the component PR [does this](https://github.com/choojs/choo/pull/639/files#diff-168726dbe96b3ce427e7fedce31bb0bcR214). Also, to prevent leaking state between pages being rendered concurrently (and consequently rendering a bloated initial state) we'd need to change `toString` to either completely overwrite `this.state` with what's passed in to `toString` or merge with the passed in state as base.

yoshuawuyts:
@tornqvist ah alright, makes sense!

PR #450, opened 2018-03-21
Enable node debugger with debug flag

tornqvist:
This is a 🙋 feature

Debugging an app running on bankai (or bankai itself, for that matter) with just `console.log` isn't the most ergonomic experience and the `debug` option wasn't doing anything anyway (one has to set the `DEBUG` env variable to enable logs). 

This adds the `debug` option which signals node to start listening for debuggers, optionally on a specific port. Though the port option isn't working for me, supposedly this is how it's intended to work (https://nodejs.org/api/process.html#process_process_debugport).

## Checklist
- [x] tests pass

## Semver Changes
Minor

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ a project setup from scratch :sparkles:.
 
   Options:
 
-    -d, --debug       output lots of logs
+    -d, --debug       start node debugger
     -h, --help        print usage
     -q, --quiet       don't output any logs
     -v, --version     print version
--- a/bin.js
+++ b/bin.js
@@ -20,7 +20,7 @@ var USAGE = `
 
   Options:
 
-    -d, --debug       output lots of logs
+    -d, --debug       start node debugger
     -h, --help        print usage
     -q, --quiet       don't output any logs
     -v, --version     print version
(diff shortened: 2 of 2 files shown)


s3ththompson:
I've been doing a lot of manual `node --inspect --inspect-brk node_modules/.bin/bankai` but for me, this would be most usable with `--inspect-brk`.

goto-bus-stop:
I like it! re ergonomics, we're supposed to send console.logs to a separate tab in the terminal ui but that is not implemented yet.

Figure people can just do `DEBUG=bankai* bankai start` if they want those logs.

goto-bus-stop:
@s3ththompson's `--inspect-brk` suggestion makes sense to me, but I don't know how that would work once the JS is already running?

tornqvist:
@s3ththompson does `--inspect-brk` make sense from a user perspective though? Seeing as it'd break in bankai source code, not in the app I am working on. We probably could inject a `debug` statement into the first line of the entry file but that'd be borderline dark arts.

s3ththompson:
I use `--inspect-brk` to give myself time to open a dedicated Node window from `chrome://inspect` when I have a debug statement in my code but the code might exit before I can get the window open (when debugging a CLI like Bankai for example, rather than a server that's continuously responding to requests). Once I'm connected I just start execution again. Maybe this is a distinction between debugging `bankai build` and `bankai start`...

https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27

tornqvist:
I hear you. When using `bankai start` I usually just resave a file after attaching a debugger to trigger a new build. That wouldn't work with `bankai build`. I'm not sure how we'd go about making that work though, as @goto-bus-stop pointed out, the process is already running.

PR #452, opened 2018-03-22
Asynchronously init choo stores before list routes

tornqvist:
This is a 🙋 feature

## Checklist
- [x] tests pass

## Context

This allows for asynchronously registering routes using choo stores. There's been a lot of talk lately about async render in choo (https://github.com/choojs/choo/pull/646) and I believe we should look into async initialization at the same time. This PR would allow us to start experimenting with that without having to commit to breaking changes in the API.

Here's an example choo store which, when registered, adds a promise to the `_experimental_prefetch` array which resolves once it has fetched all posts from some CMS and added them as routes to the app. These routes are included in static builds and enjoy proper SSR.

```javascript
module.exports = store

function store (state, emitter, app) {
  state._experimental_prefetch.push(
    getPostsFromCMS().then(function (posts) {
      posts.forEach(function (post) {
        // add route for each post
        app.route(`/posts/${post.slug}`, require('../views/post'))
      })
    })
  )
}
```

This depends on https://github.com/choojs/choo/pull/649 which allows for custom state to be passed into stores.

## Semver Changes
Minor

### DIFF
--- a/lib/graph-document.js
+++ b/lib/graph-document.js
@@ -31,16 +31,19 @@ function node (state, createEdge) {
     return self.emit('error', 'documents', entry, ssr.error)
   }
 
-  // TODO: don't pass a callback here - super hard to reason about. Find a
-  // different way instead. Perhaps a prototype with methods on it instead?
-  self.emit('ssr', { success: true, renderRoute: documentifyRoute })
-
   var fonts = extractFonts(state.assets)
-  var list = ssr.routes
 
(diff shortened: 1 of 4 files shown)


goto-bus-stop:
gotta bump this up my queue. i should get to it thursday morning!

tornqvist:
Ohh, I had forgotten about this. I'll try and resolve the conflicts sometime later this week.

PR #453, opened 2018-03-22
Fix font preload path

tornqvist:
This is a 🐛 bug fix

## Checklist
- [x] tests pass

## Context
Font preload paths were being resolved to their absolute path on disk. This fixes that by resolving their path as relative to app entry file (and optional base dir).

```diff
- <link rel="preload" as="font" crossorigin="" href="/Users/carl/site.com/assets/font.woff">
+ <link rel="preload" as="font" crossorigin="" href="/assets/font.woff">
```

## Semver Changes
Patch

### DIFF
--- a/lib/graph-document.js
+++ b/lib/graph-document.js
@@ -35,7 +35,7 @@ function node (state, createEdge) {
   // different way instead. Perhaps a prototype with methods on it instead?
   self.emit('ssr', { success: true, renderRoute: documentifyRoute })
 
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
sweet! this issue also happens for all asset paths in the FILE_LIST env variable in the service worker. Maybe it would be better to store basedir-relative asset paths in the `assets:list` node to begin with?

PR #454, opened 2018-03-28
Fix assets pathing, and add fonts to document tests

bendik:
is this a 🐛 bug fix

## Checklist
- [x] tests pass

## Context
https://github.com/choojs/bankai/pull/453

## Semver Changes
Patch

### DIFF
--- a/lib/graph-assets.js
+++ b/lib/graph-assets.js
@@ -1,5 +1,6 @@
 var trackDir = require('./track-dir')
 var utils = require('./utils')
+var path = require('path')
 
 var dirs = [
   'assets',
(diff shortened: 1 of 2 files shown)


lib/graph-assets.js
```diff
@@ -35,7 +35,9 @@ function node (state, createEdge) {
   })
 
   tracker.on('change', function () {
-    var list = tracker.list()
+    var list = tracker.list().map(function (file) {
+      return file.replace(basedir, '')
```
goto-bus-stop: would using the `path.relative` function instead also work?

bendik:
Don't know why the tests fail here.
Tried running it with same node version as CI, and it still passes on my end. And I can't quite make out what warrants the output i see on CI.
Any ideas?

bendik:
Seems to be a POSIX/windows thing with path relative https://devdocs.io/node/path#path_path_relative_from_to

Fix incoming.

goto-bus-stop:
:package: 9.10.4. thanks!

Issue #456, opened 2018-03-30
Bankai crashes when Chrome loads webpage


louiscenter:
Bankai is crashing when I initially load my choo app using Chrome. The error Bankai throws before crashing is the following:

```
Error: route '/favicon.ico' did not match.
```

This is *not* occurring on Firefox. 

## Expected Behavior

When Chrome initially loads a choo app, Bankai should gracefully handle `favicon.ico` 404ing.

## Current Behavior
Bankai crashes when Chrome requests `favicon.ico`.

## Possible Solution
I'm unsure how to resolve this issue, but I did notice this problem previously existed with Firefox. Here was a PR that resolved this last year: https://github.com/choojs/bankai/pull/218

## Your Environment
Node v8.9.0
Bankai v9.10.4

akilism:
Just wanna leave a note that this is happening in Firefox nightly as well.

goto-bus-stop:
:package: 9.10.5. Thanks!

PR #457, opened 2018-03-30
#456 Don't throw on choo SSR missing a wildcard route

bendik:
This a 🐛 bug fix

## Checklist
- [x ] tests pass

## Context
Choo SSR will do a try catch  on router matching before any rendering action is attempted

## Semver Changes
patch

### DIFF
--- a/ssr/choo.js
+++ b/ssr/choo.js
@@ -63,6 +63,15 @@ module.exports.listRoutes = function (app) {
 module.exports.render = function (app, route, cb) {
   var state = {}
   state._experimental_prefetch = []
+
+  // Use try catch here for wayfarer missing route throws
+  try {
+    app.router.match(route)
+  } catch (e) {
+    console.error(e)
+    return cb(null, { status: 404, body: 'Not found' })
+  }
+
   app.toString(route, state)
 
   // TODO: replace with p-wait-all, once it knows how to handle an empty array.
@@ -70,10 +79,9 @@ module.exports.render = function (app, route, cb) {
     .then(render, render)
 
   function render () {
-    var body = app.toString(route, state)
-    delete state._experimental_prefetch // State needs to be serializable.
     var res = { state: state }
-    if (body) res.body = body
+    delete state._experimental_prefetch // State needs to be serializable.
+    res.body = app.toString(route, state)
     if (app.state.title) res.title = app.state.title
     if (app.state.language) res.language = app.state.language
     if (app.selector) res.selector = app.selector


ssr/choo.js
```diff
@@ -63,17 +63,25 @@ module.exports.listRoutes = function (app) {
...
   Promise.all(state._experimental_prefetch)
     .then(render, render)
 
   function render () {
-    var body = app.toString(route, state)
-    delete state._experimental_prefetch // State needs to be serializable.
     var res = { state: state }
-    if (body) res.body = body
+    delete state._experimental_prefetch // State needs to be serializable.
+    res.body = app.toString(route, state)
```
goto-bus-stop: why are these lines switched?

ssr/choo.js
```diff
@@ -62,18 +62,23 @@ module.exports.listRoutes = function (app) {
...
-  Promise.all(state._experimental_prefetch)
-    .then(render, render)
+  try {
+    body = app.toString(route, state)
+    Promise.resolve().then(function () {
+      return Promise.all(state._experimental_prefetch)
+    }).then(render, render)
+  } catch (err) {
+    // logging this to make sure people can debug the whole stack trace (sorry ui)
+    console.error(err)
```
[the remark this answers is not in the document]
bendik: I'm thinking the best way would be to render the full stack trace to the browser, and otherwise keep the one-line output in the bankai ui, but couldn't quite find my way in the code to properly intercept the error.

ssr/choo.js
```diff
@@ -63,17 +63,25 @@ module.exports.listRoutes = function (app) {
...
 
-  // TODO: replace with p-wait-all, once it knows how to handle an empty array.
-  Promise.all(state._experimental_prefetch)
-    .then(render, render)
+  return new Promise(function (resolve, reject) {
+    // First pass to populate prefetch
+    try {
+      app.toString(route, state)
+    } catch (err) {
+      // logging full stack trace to aid with debugging, sorry ui.
```
bendik: is this acceptable? Was looking around to see if I could put the stacktrace in the browser, but got lost.

ssr/choo.js
```diff
@@ -63,17 +63,25 @@ module.exports.listRoutes = function (app) {
...
+      console.error(err)
```
bendik: Is this acceptable? I was looking around to see if I could put the stacktrace in the browser, but got lost. More informative than the one line the ui gives you at least.

goto-bus-stop: Hmm, I think if you call `cb()` with the stack trace wrapped in a `<code><pre>` element it might work. but the client side choo app would take over as soon as the page loads so you probably wouldn't see it for very long…

I think the best approach would be to make the SSR errors show more context over in lib/ui, I can work on that in #417

bendik: Great, I'll leave that console statement out of it then.

goto-bus-stop:
:package: 9.10.5. Thanks!

PR #468, opened 2018-04-18
Console panel

goto-bus-stop:
This is a 🙋 feature

This implements the 2nd panel for `bankai start`, showing console output from the server rendered app (don't know if there's anything else?).

SSR uses a separate Console instance, injected as a global `console` variable using [require-with-global](https://github.com/goto-bus-stop/require-with-global). That `console` writes to a stream which is saved by lib/ui.js. Pressing 2 in the `bankai start` TUI will switch to the console panel, which is an [ansi-scrollbox](https://github.com/goto-bus-stop/ansi-scrollbox) that renders the saved console output. ansi-scrollbox still needs more features and optimization (it's currently probably very slow for many thousands of lines) but it does sort of work.

## Checklist

- [ ] tests pass

## Semver Changes
Minor

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -3,7 +3,6 @@ var css = require('sheetify')
 var choo = require('choo')
 
 css('tachyons')
-console.log('whateever')
 
 var app = choo()
 
--- a/lib/ui.js
+++ b/lib/ui.js
@@ -200,7 +200,12 @@ function mainView (state) {
   }
   str += 'Server Side Rendering: ' + ssrState + '\n'
 
-  str += footer(state)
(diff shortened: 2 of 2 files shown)


yoshuawuyts:
Oh neat! - do you by any chance have any screenshots for this? it sounds pretty great!

goto-bus-stop:
now i do!

Main view

image

Log view

image

goto-bus-stop:
The "Total size" is a bit strange in the tab bar when on the log view.

On initial build, all pages are server-rendered, so you get a lot of logs at once. It might be helpful to group them somehow. Or maybe we can somehow prefix each log line with the route name that generated it!

PR #471, opened 2018-04-23
deps: add missing dependency `resolve`

jbergstroem:
This is a 🐛 bug fix.

Found while using `pnpm` - an alternative package manager that is more strict
during resolving dependency chains. Guessing it got pulled somehow using when `npm`.

We require `resolve` in [`lib/graph-document.js`](https://github.com/choojs/bankai/blob/master/lib/graph-document.js#L5) but is missing from `packages.json`.

## Checklist

- [x] tests pass

## Semver Changes
Patch or minor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "purify-css": "^1.2.6",
     "recursive-readdir": "^2.2.1",
     "recursive-watch": "^1.1.3",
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
`v9.11.1`

PR #483, opened 2018-05-16
🐛 Unify service-worker env url format

ZhouHansen:
I have to handle the urls before cache it in my `sw.js`. eg. format of '/login' and 'assets/avatar.png' is different.
so I remove the first `'/'` from '/manifest.json' and url of server-rendered route.

## Checklist
- [x] tests pass

## Context

No

## Semver Changes

Patch


goto-bus-stop:
Thanks!

(btw your master branch has different commits again, it would be easiest if you work on different branches for different patches)

PR #487, opened 2018-05-21
enable `reload` only when watch is true

ZhouHansen:
🙋 `reload` is useless when run `bankai build`, this shave about 500ms

## Checklist

- [x] tests pass

## Context
No

## Semver Changes
No

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,11 +97,16 @@ function Bankai (entry, opts) {
   })
 
   // Insert nodes into the graph.
+  var documentsDepList = [ 'assets:list', 'manifest:bundle', 'styles:bundle', 'scripts:bundle' ]
+
+  if (opts.watch !== false) {
+    documentsDepList.push('reload:bundle')
+    this.graph.node('reload', reloadNode)
+  }
   this.graph.node('assets', assetsNode)
-  this.graph.node('documents', [ 'assets:list', 'manifest:bundle', 'styles:bundle', 'scripts:bundle', 'reload:bundle' ], documentNode)
+  this.graph.node('documents', documentsDepList, documentNode)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -97,11 +97,16 @@ function Bankai (entry, opts) {
   })
 
   // Insert nodes into the graph.
+  var documentsDepList = [ 'assets:list', 'manifest:bundle', 'styles:bundle', 'scripts:bundle' ]
+
+  if (opts.watch !== false) {
```
goto-bus-stop: I think the different `opts.watch` / `metadata.reload` / `metadata.watch` checks here and below are a little confusing. Maybe it would be easier to understand if, at the start of this function, we default `opts.reload` to the value of `opts.watch` if it is not set? Then we could check `if (opts.reload)` here and `if (state.metadata.reload)` below  in graph-document.js.

goto-bus-stop:
Cool! good find :)

index.js
ZhouHansen: Yes

ZhouHansen: I agree to add an `opts.reload`, But I think it can't check totally below in graph-document.js, because document node doesn't dependent on reload node when on `bankai build` run.

goto-bus-stop: if we set `opts.reload` inside the Bankai constructor in index.js it should work I think? then you can use it in the condition on this line before adding the `reload` node. and it will be available inside the document node as `metadata.reload`. (we already pass `reload: Boolean(opts.reload)`) Or am I missing something? :thinking:

ZhouHansen: You are right, I will create a feature branch for this.

PR #492, opened 2018-05-25
check is turbolinks app

ZhouHansen:
This is a 🙋 feature.

Check whether is a turbolinks app, then add relative meta tag.

Create a `graph-package-json.js` to check it, because the document node dependents on it, if this is a turbolinks app, will add `<meta name="turbolinks-visit-control" content="reload">` to `<head>`. May combine it with `is-electron-projects.js` into one file in the future.

## Checklist

- [x] tests pass

## Context
No

## Semver Changes
minor

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ var utils = require('./lib/utils')
 var assetsNode = require('./lib/graph-assets')
 var documentNode = require('./lib/graph-document')
 var manifestNode = require('./lib/graph-manifest')
+var packageJsonNode = require('./lib/graph-package-json')
 var reloadNode = require('./lib/graph-reload')
 var scriptNode = require('./lib/graph-script')
 var serviceWorkerNode = require('./lib/graph-service-worker')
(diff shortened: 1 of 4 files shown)


goto-bus-stop:
I like the graph node for package.json! I'm not sure about this feature though … turbolinks doesn't really do anything for choo apps does it? choo handles routing and already only replaces the `<body>` parts that need to be replaced when you navigate.

ZhouHansen:
Yeah, I worked on a hybird app which is built by choo and turbolinks. Turbolinks doesn't do anything to choo, it only makes hybird app's page jumping looks more natively, it require ssr choo and that meta tag.

goto-bus-stop:
This use case is quite specific, I think it may be better solved with a custom documentify transform, like this:

```js
// package.json
{
  "devDependencies": {
    "hstream": "^1.2.0"
  },
  "documentify": {
    "transform": [
      "./add-turbolinks-meta"
    ]
  }
}
```

```js
// add-turbolinks-meta.js
var hstream = require('hstream')
module.exports = function () {
  return hstream({
    head: { _appendHtml: '<meta name="turbolinks-visit-control" content="reload">' }
  })
}
```

ZhouHansen:
Yeah, it's a better idea.

PR #493, opened 2018-05-25
enable reload when necessary, from #487

ZhouHansen:
🙋 `reload` is useless when run `bankai build`, this shave about 500ms

## Checklist

- [x] tests pass

## Context
No

## Semver Changes
minor


goto-bus-stop:
Nice! :1st_place_medal:

PR #498, opened 2018-05-30
Use `tmp` module throughout tests

goto-bus-stop:
It was already being used in some places, and is a bit less verbose.

> ## Checklist

- [x] tests pass


goto-bus-stop:
don't understand why this is so flakey…

Issue #345, opened 2017-11-27
import Hello ^ Syntax Error


jhammars:
As a beginner, it is kind of hard to figure out what syntax is supported out of the box. For example 

---
let Hello = 'Hi';
export default Hello;
---
import Hello from './components/hello';
console.log(Hello);
---

..will throw the error: "Hmmm. We're having trouble parsing a file.".

More clear docs/readme regarding what syntax to use would be awesome.

yoshuawuyts:
Oh, that's odd — I'm not getting the same error for similar code. Could you perhaps create a repro? Thanks!

jhammars:
Seems like I missed the crucial part that generated the error. Here is a rep to replicate: https://github.com/jhammars/bankai-import-hello

, where this commented line somehow is causing the error: https://github.com/jhammars/bankai-import-hello/blob/master/src/index.js#L1

Flet:
It looks like the sheetify transform does not like the `import` line.
https://github.com/choojs/bankai/blob/master/lib/graph-script.js#L53

Commenting it out the sheetify transform allows this to run.

I tried moving the sheetify transform to after the babeliy transform, but the error is still thrown. :(

Flet:
Oh interesting!
If the `// sheetify` comment is removed from here:
https://github.com/jhammars/bankai-import-hello/blob/master/src/index.js#L1

Then it works!

I think this is related to https://github.com/stackcss/sheetify/issues/131

jhammars:
Yes, It's a little peculiar. The word "sheetify" is explicitly causing the error, even when it is in a comment. Spell sheetify wrong and you are all good =)

And yes, it looks like it is related to stackcss/sheetify#131

![image](https://user-images.githubusercontent.com/26323882/33956213-2acbcc8a-e03e-11e7-9208-93dd8686be34.png)

The problem for me personally, as a JS beginner, is that I have no clue how to use sheetify in combination with ES6 module syntax.

PR #511, opened 2018-08-06
🐛Apply sheetify and glslify after babelify

real-alexei:
Looks like otherwise Babelify is applied _after_ subject transforms.

This requires https://github.com/stackcss/sheetify/pull/156 to be merged first in order to *purifycss* is able to cleanup transpiled templates code. @yoshuawuyts Please take a look at it.

### DIFF
--- a/lib/graph-script.js
+++ b/lib/graph-script.js
@@ -55,9 +55,6 @@ function node (state, createEdge) {
     })
   }
 
-  b.ignore('sheetify/insert')
-  b.transform(sheetify)
-  b.transform(glslify)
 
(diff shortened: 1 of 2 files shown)


aselivanov:
#345 should be fixed by this.

aselivanov:
@yoshuawuyts Should be good now. Please take a look. Thanks!

goto-bus-stop:
IIRC this was a potentially breaking change for some reason. We should do it though, but ideally without running sheetify and glslify as global transforms. We can remove the `global: true` from the babelify transform for the user's own code, remove the tfilter thing from it, and move it to the start.

PR #531, opened 2019-09-29
Add style tag for disabled javascript

auggod:
If browser has javascript disabled, the css will not be loaded since we use `onload="this.rel='stylesheet'` on the link tag.
I simply added a noscript tag including the link to the css.


goto-bus-stop:
thanks, looks good! it needs some test updates, still.

goto-bus-stop:
Merged in #535. Thanks!

